PR #14, opened 2021-02-15
fix link to doc

laurentheirendt:


laurentheirendt:
now as the tag has been created, this link finally works 😄

Issue #16, opened 2021-02-23
Cannot load any model


stelmo:
Hi, thanks for writing this package, I think it is a really valuable contribution to Julia's ecosystem! However, I am having some difficulty loading any model except the `Ec_core_flux1.xml` model that is used in the testing script. I have tried `iMM904.xml`, `iJO1366.xml` and `e_coli_core.xml` all recently downloaded from BIGG, see below for my code:

```julia
using SBML

# modelpath = joinpath("models", "iMM904.xml") # doesn' work
# modelpath = joinpath("models", "iJO1366.xml") # doesn't work
modelpath = joinpath("models", "e_coli_core.xml") # doesn't work
# modelpath = joinpath("models", "Ec_core_flux1.xml") # works, the model is loaded and seems functional

model = readSBML(modelpath) 
```
The error thrown is:

```julia
ERROR: 
Stacktrace:
 [1] readSBML(::String) at C:\Users\St. Elmo\.julia\packages\SBML\1vqCD\src\readsbml.jl:14
 [2] top-level scope at REPL[3]:1ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getproperty at .\Base.jl:33 [inlined]
 [2] showerror(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::SystemError) at .\errorshow.jl:128
 [3] (::Base.var"#689#690"{SystemError})(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}) at .\errorshow.jl:90
 [4] with_output_color(::Function, ::Symbol, ::IOContext{Base.TTY}; bold::Bool) at .\util.jl:72
 [5] with_output_color(::Function, ::Symbol, ::IOContext{Base.TTY}) at .\util.jl:70
 [6] showerror(::IOContext{Base.TTY}, ::SystemError, ::Array{Base.StackTraces.StackFrame,1}; backtrace::Bool) at .\errorshow.jl:89
 [7] showerror at .\errorshow.jl:88 [inlined]
 [8] display_repl_error(::Base.TTY, ::SystemError, ::Array{Union{Ptr{Nothing}, Base.InterpreterIP},1}) at c:\Users\St. Elmo\.vscode\extensions\julialang.language-julia-1.1.27\scripts\packages\VSCodeServer\src\repl.jl:124
[...]

PR #18, opened 2021-02-23
Fix problems for reactions that don't have kinetic laws associated

exaexa:
Refs #16 

PS. it would be nice to bump version after this.


laurentheirendt:
Is there any test for this?

exaexa:
Travis is taking the old SBML_jll version again. Was there some trick to speed it up last time?

exaexa:
Both SBML_jll and extra tests are fixed now.

laurentheirendt (approved):

laurentheirendt:
maybe we should bump the version - can you do? thx

Issue #12, opened 2021-02-11
Sparse matrix output


exaexa:
`getS` currently returns a dense matrix (that can become quite inconvenient even for modestly sized models). Return some version of a sparse matrix.

exaexa:
Closed by #19

PR #19, opened 2021-02-26
Support sparse S matrix output

exaexa:
This should allow loading of pretty big models.

Version bump is necessary (API has changed slightly).


exaexa:
Fixes #12

laurentheirendt:
Can you add a test for the sparse matrix output?

laurentheirendt (approved):

PR #24, opened 2021-03-14
load more useful metadata

### LINKED ISSUE #22: readSBML extension
Hi, I'm in the early stages of building a Julia version of the CobraTool box, see [here](https://github.com/stelmo/CobraTools.jl) for reference. For this I need an SBML reader/writer, and hence my interest in this package. In my package I essentially adopt the COBRApy framework of storing a model as an array of genes, metabolites and reactions. For each of these I aim to import most of the data contained in the model source file, e.g. for metabolites this includes (per metabolite):

1. id
2. name
3. compartment
4. formula
5. charge
6. notes
7. annotations

This is documented in the `Model structure` section of the documentation of CobraTools.jl if you are interested (although the docs are a work in progress).

Currently in SBML.jl only the first three fields from the model source file are being imported. I have forked the package and added some more fields to make SBML.jl mesh better with my package (currently only for species/metabolites). Would you mind looking at the modifications on my fork and letting me know if you would be open to pulling them onto SBML.jl (once I've extended the genes and reactions, and added some tests of course)?

exaexa:
work-in-progress now, opened for tracking.

Closes #22 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SBML"
 uuid = "e5567a89-2604-4b09-9718-f5f78e97c3bb"
 authors = ["Mirek Kratochvil <miroslav.kratochvil@uni.lu>", "LCSB R3 team <lcsb-r3@uni.lu>"]
-version = "0.2.0"
+version = "0.3.0"
 
 [deps]
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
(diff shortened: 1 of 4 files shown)


exaexa:
Depends on #23 now (this is using some of the code).

Otherwise I guess this is ready to merge. Minor version is bumped to 0.3 so that people don't have hard time depending on this.

laurentheirendt:
can you add some documentation to the different functions?

laurentheirendt (changes requested):
can you add some documentation to the different functions?

exaexa:
The new functions are internal and kinda straightforward so I decided not to care, will add.

laurentheirendt (approved):

Issue #31, opened 2021-03-31
SBML.jl and SciML integration


anandijain:
Hello, I've been working with @paulflang on building an SBML parser that would convert SBML ODEs into ModelingToolkit.jl form, to be used with the SciML ecosystem.

Here is the repo https://github.com/anandijain/SBML.jl/
It would not make sense for us to double up work.

If any maintainers are on slack, we have a chat for building biological model importers (SBML, CellML, bionetgen) into SciML. 
It'd be good to see how we can help y'all and still get tight integration with SciML.

laurentheirendt:
Hello @anandijain and @paulflang 😄 

Good to see your efforts! The main reason behind using `SBML_jll` and [libSBML](http://sbml.org/Software/libSBML) is that it is a well-established library and it was the most straightforward way to interface to that library from Julia without having to resort to writing a parser from scratch.

For our use case, @exaexa and I thought it would not make too much sense to implement a full Julia parser from scratch - of course, this is always welcome, but I doubt we would see any major benefits performance wise.

What I could imagine is to actually support both, a native `SBML.jl` package and a `libSBML.jl` package (that would be this one). We should also have a chat with @skeating or @fbergmann from the SBML team to get their input 👍🏼

exaexa:
I originally thought that writing the XML parser from below would be viable too, but given all the complexity in SBML specification this was not the way to go. Technically, it would be best to have the SBML schema published in a machine-accessible format and automatically generate the XML parser from it -- there have been certainly some efforts on writing the schema, but I'm not at all sure how that's going.

paulflang:
Great suggestion to hear Sarah Keating's and Frank Bergmann’s thoughts. Would also be interesting to hear were we stand with the schema.

Since you are more interested in FBA and we are more interested in ODEs and SSA, what we do with the SBML model once it is imported into Julia is different. But I see you have a `Model` type that is somewhat similar to our `SbmlModel` type. I could imagine that defining a common type would make sense (no matter if we use Julia native parsers or not, and no matter what we do downstream with the model). But I am not familiar with how FBA models are specified in SBML. So I will have to read up on that, as well as having a look at your code.

exaexa:
> But I see you have a Model type that is somewhat similar to our SbmlModel type. I could imagine that defining a common type would make sense (no matter if we use Julia native parsers or not, and no matter what we do downstream with the model).

[...]

PR #32, opened 2021-04-07
Draft for converting `Model` to `ModelingToolkit` types

paulflang:
This is not working code, but just a rough idea how `Model` can be adapted and further converted to the `ModelingToolkit` types we need. I think we still have to make several more abstract design choices, but perhaps this concrete example helps to get an idea of what is needed.

### DIFF
--- a/src/structs.jl
+++ b/src/structs.jl
@@ -63,6 +63,7 @@ bounds (in tuples `lb` and `ub`, with unit names), and objective coefficient
 (`oc`). Also may contains `notes` and `annotation`.
 """
 struct Reaction
+    mathml::String
     stoichiometry::Dict{String,Float64}
     lb::Tuple{Float64,String}
     ub::Tuple{Float64,String}
@@ -108,8 +109,8 @@ objects.
 struct Model
     parameters::Dict{String,Float64}
     units::Dict{String,Vector{UnitPart}}
-    compartments::Vector{String}
-    species::Dict{String,Species}
+    compartments::Dict{String,Float64}  # PL: Float64 to describe size
+    species::Dict{String,Tuple{Species,Float64}}  # PL: Tuple of Species and initialAmount
     reactions::Dict{String,Reaction}
     gene_products::Dict{String,GeneProduct}
     notes::Maybe{String}
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -1,5 +1,9 @@
 create_var(x) = Num(Variable(Symbol(x)))
-create_param(x) = Num(Sym{ModelingToolkit.Parameter{Real}}(Symbol(x)))
+function create_param(x)
+  p = Sym{Real}(Symbol(x))
+  ModelingToolkit.toparam(p)
+  p
+end
 
 
 """ ModelingToolkit.Reaction constructor """
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -1,0 +1,38 @@
+create_var(x) = Num(Variable(Symbol(x)))
+create_param(x) = Num(Sym{ModelingToolkit.Parameter{Real}}(Symbol(x)))
+
+
+""" ModelingToolkit.Reaction constructor """
+function ModelingToolkit.Reaction(reaction::Reaction; only_use_rate=true)
+    # pass
+end
+
+
+""" ReactionSystem constructor """
+function ModelingToolkit.ReactionSystem(model::Model)
+    rxs = [Reaction(reac...; only_use_rate=true) for reac in values(model.reactions)]
+    t = Variable(:t)
+    species = [create_var(spec) for spec in keys(model.species)]
+    pars = [create_param(p) for p in keys(model.parameters)]
+    comps = [create_param(c) for c in keys(model.compartments)]
+    pc = vcat(pars,comps)
+    ReactionSystem(rxs,t,species,pc)
+end
+
+
+""" ODESystem constructor """
+function ModelingToolkit.ODESystem(model::Model)
+    rs = ReactionSystem(model)
+    u0map = [create_var(k) => v[2] for (k, v) in model.species]
+    parammap = vcat([create_param(k) => v for (k,v) in model.parameters],
+                    [create_param(k) => v for (k,v) in model.compartments])
+    defaults = vcat(u0map, parammap)
+    convert(ODESystem, rs, defaults=defaults)
+end
+
+
+""" ODEProblem constructor """
+function ModelingToolkit.ODEProblem(model::Model,tspan)  # PL: Todo: add u0 and parameters argument
+    odesys = ODESystem(model)
+    ODEProblem(odesys, [], tspan)
+end
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -7,7 +7,7 @@ const VPtr = Ptr{Cvoid}
 Read the SBML from a XML file in `fn` and return the contained `Model`.
 """
 function readSBML(fn::String)::Model
-    doc = ccall(sbml(:readSBML), VPtr, (Cstring,), fn)
+    doc = ccall(sbml(:readSBML), VPtr, (Cstring,), fn)  # PL: We need to promote local parameters. We may need to convert every species to extensive units.
     try
         n_errs = ccall(sbml(:SBMLDocument_getNumErrors), Cuint, (VPtr,), doc)
         for i = 0:n_errs-1
@@ -203,6 +203,7 @@ function extractModel(mdl::VPtr)::Model
 
         # kinetic laws store a second version of the bounds and objectives
         kl = ccall(sbml(:Reaction_getKineticLaw), VPtr, (VPtr,), re)
+        mathml = ccall(sbml(:KineticLaw_getMath), VPtr, (VPtr,), kl)  # PL: get MathML of kineticLaw; may need further conversion to XML string or EzXML node before converting to a symbolic formula via MathML.jl unless a future version of MathML accepts VPtr arguments.
         if kl != C_NULL
             for j = 1:ccall(sbml(:KineticLaw_getNumParameters), Cuint, (VPtr,), kl)
                 p = ccall(sbml(:KineticLaw_getParameter), VPtr, (VPtr, Cuint), kl, j - 1)
@@ -279,6 +280,7 @@ function extractModel(mdl::VPtr)::Model
 
         reid = unsafe_string(ccall(sbml(:Reaction_getId), Cstring, (VPtr,), re))
         reactions[reid] = Reaction(
+            mathml,
             stoi,
             lb,
             ub,
--- a/src/stuct_suggestions.jl
+++ b/src/stuct_suggestions.jl
@@ -1,0 +1,79 @@
+### This file suggest to possibilities of adapting the types to accommodate also dynamic models
+
+############# Version 1 #################
+abstract type Reaction end
+abstract type Species end
+abstract type SbmlModel end
+
+struct DynamicSpecies <: Species
+    initialAmout::Float64
+    # etc
+end
+
+struct FbaSpecies <: Species
+    # etc
+end
+
+struct DynamicReaction <: Reaction
+    kineticLaw::String
+    # etc
+end
+
+struct FbaReaction <: Reaction
+    lb::Float64
+    ub::Float64
+    # etc
+end
+
+struct DynamicReactionModel <: SbmlModel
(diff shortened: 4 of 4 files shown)


anandijain:
@laurentheirendt if you have a sec to peek this and perhaps https://github.com/SciML/ModelingToolkit.jl/blob/master/test/reactionsystem.jl and https://github.com/SciML/Catalyst.jl/issues/312 it would be good to have some context supporting tomorrows meeting.
Thanks!

src/structs.jl
```diff
@@ -63,6 +63,7 @@ bounds (in tuples `lb` and `ub`, with unit names), and objective coefficient
 (`oc`). Also may contains `notes` and `annotation`.
 """
 struct Reaction
+    mathml::String
```
anandijain, line 66: Maybe use XML node if possible? I suppose it might not work

exaexa:
Hi everyone! Do we have some example SBML that would contain all these features?

Also, if I get it correctly, the aim here is to extract:
- the formula that's specified in MathML
- initialAmount for species
- compartment sizes
...anything else?

paulflang:
Hi @exaexa . [This](https://github.com/anandijain/SBML.jl/blob/main/test/data/case01.xml) would be a simplistic example containing a `rateLaw`, `initialAmount` and compartment `size`.

After yesterday's meeting we decided to use different concrete types do describe `Fba` and `DynamicReactionModels` (all subtyping the same abstract model, for instance called `SbmlModel`; we did not decide on precise namings yet). Later I realised the following (copying over from Slack):

> If we create different types for `FbaModel` and `DynamicReactionModel` to accommodate that in dynamics models `Species` need a `initialAmount`, and `Reactions` need a `kineticLaw` (but no upper and lower bound), we should probably also create different `Species` types (`FbaSpecies` and `DynamicSpecies`) and `Reaction` types (`FbaReaction` and `DynamicReaction`).
> Alternatively, we could just use the same types for FBA and dynamic models/species/reactions. This would mean few more fields of type `Maybe` (where `const Maybe{X} = Union{Nothing,X}`). It would also mean that a downstream program would not know if a Model is FBA or dynamic. Or we could do hybrid, where the distinction is only on the Model level.
> Either way, the API would stay the same `m = readSBML("myModel.xml")` .

What are your thoughts on that?

Also we need to figure out how to solve extensive/intensive unit problem as discussed in #31 today.

There are other things that can be specified in SBML (mass conservation equations and quasi steady-state approximations), but for now the three points you mentioned are what we would need.

exaexa:
OK great, so the list expands to this:

- the kinetic law formula that's specified in MathML (side note: what does it mean if there's a compartment written in the equation?)
- all parameters + possibly notes
- initialAmount for species
- compartment metadata, incl. sizes
- multiple subtypes of reactions and species (this can be handled quite easily using some variant type on the reaction list, would you have a sample with `DynamicReaction` etc?)
- unit checks for everything, possibly `hasOnlySubstanceUnits` handling

Technical:
- do we want to reparse the MathML? SBML gives some access to the DOM of it, which I can pull to julia as a structure.
- what's `FbaReaction` and `DynamicReactions`? (preferably, are there examples?)

paulflang:
> what does it mean if there's a compartment written in the equation?

The units of reactions have to be in amount per time, not amount per volume per time.

anandijain:
image

From the spec. I think case 1 will be the fastest to get working for our two cases (FBA and ODE).

But perhaps the "correct" way is strictly following the image, where specific subtypes require not a Maybe but a concrete type.

paulflang:
@anandijain : so are you suggesting that all the things in the image should become abstract types in the long run. and each of them might have different concrete types?

For instance
```
abstract type Species end

struct FbaSpecies <: Species
    ...
end

struct DynamicSpecies <: Species
    ...
end
```

anandijain:
I was thinking, where T is just a pointer to the XML or C++ struct. Because we really want to avoid just building a duplicate copy of the C++ stuff

```julia
# all Option types
struct SubModel <: Model 
    a::Union{Nothing, T}
    b::Union{Nothing, T}
    #...
end
```
I'm not really sure this is the best design, it just to me looks the most similar to that snippet from the spec

paulflang:
If all the fiels are `Union{Nothing, T}`, then we do no longer need different types anymore, right. So we could just do:
```Julia
struct Model 
    a::Union{Nothing, T}
    b::Union{Nothing, T}
    #...
end
```
But then we have the problem that when we look at an instance of `Model` we do not know if it is FBA, dynamic or whatever.

anandijain:
>do we want to reparse the MathML? SBML gives some access to the DOM of it, which I can pull to julia as a structure.

@exaexa have you glanced at the readme for https://github.com/SciML/MathML.jl ? Essentially we would want a way to get back equations/variables in https://github.com/JuliaSymbolics/Symbolics.jl types

anandijain:
> If all the fields are `Union{Nothing, T}`, then we do no longer need different types anymore, right. So we could just do:
> 
> ```julia
> struct Model 
>     a::Union{Nothing, T}
>     b::Union{Nothing, T}
>     #...
> end
> ```
> 
> But then we have the problem that when we look at an instance of `Model` we do not know if it is FBA, dynamic or whatever.

I think we could make subtypes and design the interface so that 
a subtype `FbaModel <: Model` must implement these methods for it to satisfy being an FbaModel. 

I think a similar way could be done for convert(ODEsystem, Model) or convert(ReactionSys, Model)

paulflang:
Sorry, what does it mean for a type to implement methods? I thought types implement no other methods than their constructors. But methods can be constrained to only accept certain types. Where am I wrong? And what does "these" refer to in

> a subtype FbaModel <: Model must implement these methods

paulflang:
also, the next problem we face is that in `a::Union{Nothing,T}` `a` might be of different type in an FBA model compared to a dynamic model. To give you a concrete example, if `a` is a `Reaction`, then `Reactions` in FBA models contain upper and lower bounds, whereas `Reactions` in dynamic models contain a kineticLaw instead.

anandijain:
im treating all T as ptr to a node of s expr. if this is the wrong way to think about it lmk though

exaexa:
I suggest not worrying about precise types of everything now, I'll see what's gonna be best when implementing it. The model in SBML.jl should be descriptive (in the libsbml sense), although incomplete; at the same time accessible programatically from Julia. We can make functions that "simplify" the loaded model to domain-specific data subset, like the `getS` we have for constraint-based models. That should hopefully solve the problem with Fba* vs Dynamic* type variants. There is no point in making actual subtypes here, as there are (as far as I get it) no functions that would need to transparently work on both variants.

@anandijain I guess I can convert stuff from libsbml representation directly into Symbolic without much trouble, would that work for you? Technically I'd try to do this using some parts of MathML.jl, likely patching it a bit in the process.

exaexa:
> im treating all T as ptr to a node of s expr. if this is the wrong way to think about it lmk though

well the purpose of SBML.jl is to prevent the complexity and dangers of handling anything pointy, so I guess we're not going for actual pointers.

Regarding your code, `T` there seems a bit like a type variable (aka template type aka type parameter, depending on what type language you like). For julia specifics, it is missing the `where {T ...}` part that would specify what it can become.

anandijain:
> @anandijain I guess I can convert stuff from libsbml representation directly into Symbolic without much trouble, would that work for you? Technically I'd try to do this using some parts of MathML.jl, likely patching it a bit in the process.

amazing!

paulflang:
Sounds great! So the kineticLaw will already be Symbolic, i.e. contain parameters and variables as Symbolic (of type `Num` I believe). `Model.parameter` is currently of type `Dict{String,Float64}`. Would it make sense to make it `Dict{Num,Float64}`, then (analogous for `Model.species`)?

laurentheirendt:
I am not sure if the prefix `FBA` is appropriate. Along the lines of `DynamicReactionModel`, we could think of `SystemsBiologyModel`. What do you think?

paulflang:
The way I understood @exaexa is that he won't implement subtypes of `Model`.  

But `SystemsBiology` would include every FBA, Dynamic Models, Boolean Models (of which I do not even know if they can be soecified in SBML). So I think that would not be the right name.

laurentheirendt:
Maybe then `ConstraintBasedModel` would fit best

exaexa:
@paulflang I just noticed that there's also a `initialConcentration` property supported by SBML. Is that used? (It would clarify a bit of things for me, unfortunately now the SBML specification is completely opaque for me in this direction).

paulflang:
`initialConcentration` is frequently used, but not in the model I've sent you. I find page 50 last paragraph of [this document useful](http://sbml.org/Special/specifications/sbml-level-3/version-2/core/release-2/sbml-level-3-version-2-release-2-core.pdf).

But we want to use extensive quantities throughout `ReactionSystem`.

- So if `initialConcentration` is used, we will have to multiply with compartment `size` to get the amount.
- If a species `hasOnlySubstanceUnits=false`, we will have to modify all occurences of this species (e.g. in `kineticLaw`s) accordingly (still waiting for a response on #31 to clarify this).

exaexa:
I'm still completely confused about the semantics there, esp. the meaning of `hasOnlySubstanceUnits`. :sweat_smile: 

There are apparently 5 possibilities for the species
- `initialConcentration` is set, hOSU is `false`
- `initialConcentration` is set, hOSU is `true`
- `initialAmount` is set, hOSU is `false`
- `initialAmount` is set, hOSU is `true`
- nothing of the above is set

are there any other "states" of the species to represent?

paulflang:
So am I. In SBML Level 2 there is also the possibility of having either initialConcentration or initialAmount and hOSU just not there:
```
<species metaid="COPASI2" id="A" name="A" compartment="compartment" initialConcentration="1" boundaryCondition="false" constant="false">
```

But perhaps we should forget about all other levels than 3 for now.

We can also raise a not implemented error if `boundaryCondition` or `constant` are true, but in the long run we'd probably want to have these attributes (if I remember correctly, they indicate if species shall not be consumed or produced even if they appear in reactions).

We also need the `compartment` attribute to calculate `initialAmounts` in case only `initialConcentrations` are given.

exaexa:
> So am I.
Ah, semantic chaos!!! :grinning: 

> In SBML Level 2 there is also the possibility of having either initialConcentration or initialAmount and hOSU just not there:
> 
> ```
> <species metaid="COPASI2" id="A" name="A" compartment="compartment" initialConcentration="1" boundaryCondition="false" constant="false">
> ```
> 
> But perhaps we should forget about all other levels than 3 for now.

Also, I just noticed that the spec in SBML docs does say that if the thing is concentration based, hOSU should be _set_, but doesn't specify to what value it should be set. :D

> 
> We can also raise a not implemented error if `boundaryCondition` or `constant` are true, but in the long run we'd probably want to have these attributes (if I remember correctly, they indicate if species shall not be consumed or produced even if they appear in reactions).

Technically, in SBML I can amass whatever number of flags that I can parse out reliably (SBML.jl is about data representation, not about interpretation). I should add `constant` because that really looks most useful now.

> 
> We also need the `compartment` attribute to calculate `initialAmounts` in case only `initialConcentrations` are given.

Compartment ID is already there, I'll check if compartments have sizes.

paulflang:
> Also, I just noticed that the spec in SBML docs does say that if the thing is concentration based, hOSU should be set, but doesn't specify to what value it should be set.

Can you ping Sarah or Frank with this (I cannot mention them here in this thread)?

> I should add `constant` because that really looks most useful now.

Maybe add `boundaryConditions` as well? Then its done for good.

anandijain:
>Can you ping Sarah or Frank with this (I cannot mention them here in this thread)? 

Yes you can. It only recommends contributors though.

@fbergmann @skeating Maybe you could provide a little context for this? We would greatly appreciate it!

exaexa:
News: compartments and everything else should be in #39 .

fbergmann:
> > Also, I just noticed that the spec in SBML docs does say that if the thing is concentration based, hOSU should be set, but doesn't specify to what value it should be set.
> 
> Can you ping Sarah or Frank with this (I cannot mention them here in this thread)?
> 
> > I should add `constant` because that really looks most useful now.
> 
> Maybe add `boundaryConditions` as well? Then its done for good.

Indeed these attributes all had default values in SBML Level 2, in SBML Level 3 these had been removed. I would use for concentration based species: 

```
       initialConcentration="1" 
       substanceUnits="substance"  <-- replace with what specific unit it was
       hasOnlySubstanceUnits="false"  
       boundaryCondition="false" 

```

i hope this helps

laurentheirendt:
closing in favor of #40

PR #36, opened 2021-04-13
Update README.md

anandijain:
make explicit that pkg is registered in readme


laurentheirendt:
thanks @anandijain 👍🏼

PR #37, opened 2021-04-13
Sanitize string handling, make species' names optional

exaexa:
Should improve upon #35.

PR #39, opened 2021-04-13
Extract kinetics-relevant information

exaexa:
Built upon `mk-strings` in #37 (would be nice to merge to the same release).

Generally this is a first response to #31 and #32, likely not complete yet.

Cc @paulflang

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SBML"
 uuid = "e5567a89-2604-4b09-9718-f5f78e97c3bb"
 authors = ["Mirek Kratochvil <miroslav.kratochvil@uni.lu>", "LCSB R3 team <lcsb-r3@uni.lu>"]
-version = "0.3.2"
+version = "0.4.0"
 
 [deps]
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
--- a/src/SBML.jl
+++ b/src/SBML.jl
@@ -6,6 +6,7 @@ using SparseArrays
 include("structs.jl")
 include("version.jl")
 include("readsbml.jl")
+include("math.jl")
 include("utils.jl")
 
 sbml = (sym::Symbol) -> dlsym(SBML_jll.libsbml_handle, sym)
--- a/src/math.jl
+++ b/src/math.jl
@@ -1,11 +1,40 @@
 
 """
-    parseMath(ast::VPtr)::Math
+    ast_is(ast::VPtr, what::Symbol)::Bool
+
+Helper for quickly recognizing kinds of ASTs
+"""
+ast_is(ast::VPtr, what::Symbol)::Bool = ccall(sbml(what), Cint, (VPtr,), ast) != 0
+
+"""
+    parse_math(ast::VPtr)::Math
 
 This attempts to parse out a decent Julia-esque ([`Math`](@ref) AST from a
 pointer to `ASTNode_t`.
 """
-function parseMath(ast::VPtr)::Math
-    @info "got $ast"
-    MathIdent("placeholder")
+function parse_math(ast::VPtr)::Math
+    if ast_is(ast, :ASTNode_isName) || ast_is(ast, :ASTNode_isConstant)
+        return MathIdent(get_string(ast, :ASTNode_getName))
+    elseif ast_is(ast, :ASTNode_isConstantNumber)
+        return MathVal(ccall_sbml(:ASTNode_getValue), Cdouble, (VPtr,), ast)
+    elseif ast_is(ast, :ASTNode_isFunction)
+        return MathApply(
+            get_string(ast, :ASTNode_getName),
+            [
+                parse_math(ccall(sbml(:ASTNode_getChild), VPtr, (VPtr, Cuint), ast, i - 1))
+                for i = 1:ccall(sbml(:ASTNode_getNumChildren), Cuint, (VPtr,), ast)
+            ],
+        )
+    elseif ast_is(ast, :ASTNode_isOperator)
+        return MathApply(
+            string(Char(ccall(sbml(:ASTNode_getCharacter), Cchar, (VPtr,), ast))),
+            [
+                parse_math(ccall(sbml(:ASTNode_getChild), VPtr, (VPtr, Cuint), ast, i - 1))
+                for i = 1:ccall(sbml(:ASTNode_getNumChildren), Cuint, (VPtr,), ast)
+            ],
+        )
+    else
+        @warn "unsupported math element at $ast"
+        return MathIdent("?unsupported?")
+    end
 end
--- a/src/math.jl
+++ b/src/math.jl
@@ -1,0 +1,11 @@
+
+"""
+    parseMath(ast::VPtr)::Math
+
+This attempts to parse out a decent Julia-esque ([`Math`](@ref) AST from a
+pointer to `ASTNode_t`.
+"""
+function parseMath(ast::VPtr)::Math
+    @info "got $ast"
+    MathIdent("placeholder")
+end
--- a/src/math.jl
+++ b/src/math.jl
@@ -7,6 +7,16 @@ Helper for quickly recognizing kinds of ASTs
 ast_is(ast::VPtr, what::Symbol)::Bool = ccall(sbml(what), Cint, (VPtr,), ast) != 0
 
 """
+    parse_math_children(ast::VPtr)::Vector{Math}
+
+Recursively parse all children of an AST node.
+"""
+parse_math_children(ast::VPtr)::Vector{Math} = [
+    parse_math(ccall(sbml(:ASTNode_getChild), VPtr, (VPtr, Cuint), ast, i - 1)) for
+    i = 1:ccall(sbml(:ASTNode_getNumChildren), Cuint, (VPtr,), ast)
+]
+
+"""
     parse_math(ast::VPtr)::Math
 
 This attempts to parse out a decent Julia-esque ([`Math`](@ref) AST from a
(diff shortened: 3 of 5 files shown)


exaexa:
So this now parses some basic MathML, the resulting AST is super lightweight, and can be
- easily transformed to whatever else AST there is
- extended with supporting other MathML functionality (mostly lambdas now).

Demo:
```
julia> readSBML("00870-sbml-l3v1.xml").reactions["fasterReaction"].kinetic_math
SBML.MathApply("*", SBML.Math[
  SBML.MathApply("+", SBML.Math[
    SBML.MathApply("*", SBML.Math[
      SBML.MathIdent("Kf1"),
      SBML.MathIdent("A4")]),
    SBML.MathApply("-", SBML.Math[
      SBML.MathApply("*", SBML.Math[
        SBML.MathIdent("Kr1"),
        SBML.MathIdent("A2")])])]),              # spoiler: I indented this manually
  SBML.MathIdent("C")])

julia> readSBML("00870-sbml-l3v1.xml").species["A1"].initial_amount
(2.0, "substance")
```

@anandijain @paulflang would this work as a base for future fun with math?

anandijain:
I notice that MathML is not added as a dependency, but you construct your own types. Is there any way we can upstream that? if not, totally fine, but it looks close.

I'm wondering why the ast can't just be one big string, ie is there not a function to render the whole section within a mathml namespace as a raw mathml string, then we can just do the hand-off

anandijain:
As it just says MathIndent, I can't tell what is a parameter and what is a variable

anandijain:
> As it just says MathIndent, I can't tell what is a parameter and what is a variable

I guess we just need to match it from the other part of the Reaction, which I think specifies this

paulflang:
So here is what I would have thought of:

1. `LCSB-BioCore/SBML.Reaction` has a `kineticLaw` field that stores an XML string.
2. `MathML.parse_node()` is used to parse the XML string and return the symbolic expression of it (`Num(Variable(:X))` symbols connected by valid Julia operators like `+`, `*`, etc.).
3. When converting an `LCSB-BioCore/SBML.Model` to a `ReactionSystem` I do `substitute(kineticLaw, substitutiondict)` as is done in `CellMLToolkit` and `anandijain/SBML.jl` to convert the symbols to the right types depending on whether they are parameters or variables.
4. If any of the variables has `hOSU=false` I need to correct for `size` of the corresponding compartment.

But I could imagine adding a method to `MathML.parse_node()` that supports `LCSB-BioCore/SBML.MathApply` as argument type. @anandijain and @exaexa : Do you see any advantages in that

exaexa:
> I notice that MathML is not added as a dependency, but you construct your own types. Is there any way we can upstream that? if not, totally fine, but it looks close.

I would love that, but turns out (!) that SBML math minilanguage isn't really 100% MathML... Certainly the structured way to access it is the SBML AST, which is not MathML at all...

> 
> I'm wondering why the ast can't just be one big string, ie is there not a function to render the whole section within a mathml namespace as a raw mathml string, then we can just do the hand-off

You'd need to somehow convert back it to XML, not sure if it's viable this way.

> As it just says MathIndent, I can't tell what is a parameter and what is a variable

The identifiers are _not_ distinguished in MathML or SBML, they are just in `<ci>...</ci>`. SBML does some effort in distinguishing named constants, but I'm not sure if it helps much. SBML defines some identifier scoping rules (short version: all identifiers should be distinct to avoid any ambiguity, except you can shadow this big unambiguous set of all identifiers by LocalParameters).

> But I could imagine adding a method to MathML.parse_node() that supports LCSB-BioCore/SBML.MathApply as argument type. @anandijain and @exaexa : Do you see any advantages in that

This minilanguage with `Math*` types represents what SBML "sees" in the data. Writing conversion functions from it to any other type is pretty much straightforward. I'm working on reaction defs now, will make a demo either for MathML or for Symbolics right after that. (Notably, MathML is probably a bit too general-- it seems to be modeled after s-expressions and (indirectly) allows partial application of functions, which is honestly the last thing you'd ever want to reliably translate to ODEs :grinning: )

anandijain:
`<ci>` is how we define symbolic variables though? 

> This minilanguage with Math* types represents what SBML "sees" in the data. Writing conversion functions from it to any other type is pretty much straightforward. I'm working on reaction defs now, will make a demo either for MathML or for Symbolics right after that.

Thanks for this clarification. 

> Notably, MathML is probably a bit too general-- it seems to be modeled after s-expressions and (indirectly) allows partial application of functions, which is honestly the last thing you'd ever want to reliably translate to ODEs 😀

Could you expand on this part? What is an example of partial application that would cause trouble for ODE lowering?
I will have a better sense of what you mean once I actually read through your code 😆, so feel free to ignore this.

anandijain:
>I would love that, but turns out (!) that SBML math minilanguage isn't really 100% MathML... Certainly the structured way to access it is the SBML AST, which is not MathML at all... 

Sorry for the spam, ignore that previous one. I reread this and it makes a little more sense. I'll have to play around with this minilanguage AST to get a feel for it. 

Is there no way to get the raw MathML XML string?

paulflang:
> <ci> is how we define symbolic variables though? (in MathML)

Well as far as I understand this is how you defined symbols in MathML. But MathML is agnostic of weather that is a MTK.parameter (SBML parameter) or MTK.variable (SBML species). But the program that calls `MathML.parse_node()` can convert the output to appropriate MTK types using the `substitute()` function.

exaexa:
> > Notably, MathML is probably a bit too general-- it seems to be modeled after s-expressions and (indirectly) allows partial application of functions, which is honestly the last thing you'd ever want to reliably translate to ODEs grinning
> 
> Could you expand on this part? What is an example of partial application that would cause trouble for ODE lowering?
> I will have a better sense of what you mean once I actually read through your code laughing, so feel free to ignore this.

Technically-- (opening pandora's box of language design:) it supports lambdas, thus it supports the Y combinator, and if you translated that to ODE, you could very easily force your ODE system to simulate a full turing machine... That is very likely not what you want, ODEs are sufficiently hard all by themselves :D Ever worse, this can be done without actually using lambdas: the "global" FunctionDefinitions let you you define a SKI base calculus, then you can implement the recursion without actually writing a lambda or anything that would seem even remotely recursive.

From the other direction, there are the "special" SBML identifier names (aka constants), the handling of which is still fishy for me.

> Is there no way to get the raw MathML XML string?

Not sure, but I didn't find anything good in here: http://sbml.org/Software/libSBML/docs/c-api/class_a_s_t_node__t.html

Perhaps getting the libxml handle out of SBML and re-encoding it using libxml functions could work?

exaexa:
Cc @stelmo -- this contains kinetics, might be interesting for your usecases.

paulflang:
> all identifiers should be distinct to avoid any ambiguity, except you can shadow this big unambiguous set of all identifiers by LocalParameters

On that note, would it be possible to (optionally) use the [SBMLLocalParameterConverter](https://model.caltech.edu/software/libsbml/5.18.0/docs/formatted/cpp-api/class_s_b_m_l_local_parameter_converter.html) and perhaps also the [SBMLInitialAssingmentConverter](https://model.caltech.edu/software/libsbml/5.18.0/docs/formatted/cpp-api/class_s_b_m_l_initial_assignment_converter.html)?

paulflang:
Unfortunately, bidirectional reactions in SBML have only one `kineticLaw` and I cannot find a converter that disentangles the forward and reverse rate. `ReactionSystem` would separate forward and reverse `kineticLaws` though. Consider for example the reaction
`A <-> B` below. Is there a good way of splitting the `kineticLaw` in a forward and reverse `kineticLaw`?
```
     <reaction metaid="COPASI4" id="reaction" name="reaction" reversible="true" fast="false">
        <annotation>
          <COPASI xmlns="http://www.copasi.org/static/sbml">
            <rdf:RDF xmlns:dcterms="http://purl.org/dc/terms/" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
              <rdf:Description rdf:about="#COPASI4">
                <dcterms:created>
                  <rdf:Description>
                    <dcterms:W3CDTF>2021-04-14T11:02:50Z</dcterms:W3CDTF>
                  </rdf:Description>
                </dcterms:created>
              </rdf:Description>
            </rdf:RDF>
          </COPASI>
        </annotation>
        <listOfReactants>
          <speciesReference species="A" stoichiometry="1" constant="true"/>
        </listOfReactants>
        <listOfProducts>
          <speciesReference species="B" stoichiometry="1" constant="true"/>
        </listOfProducts>
        <kineticLaw>
          <math xmlns="http://www.w3.org/1998/Math/MathML">
            <apply>
              <times/>
              <ci> compartment </ci>
              <apply>
                <minus/>
                <apply>
                  <times/>
                  <ci> k1 </ci>
                  <ci> A </ci>
                </apply>
                <apply>
                  <times/>
                  <ci> k2 </ci>
                  <ci> B </ci>
                </apply>
              </apply>
            </apply>
          </math>
          <listOfLocalParameters>
            <localParameter id="k1" name="k1" value="0.1"/>
            <localParameter id="k2" name="k2" value="0.1"/>
          </listOfLocalParameters>
        </kineticLaw>
      </reaction>
```

paulflang:
The math above would be `compartment*(k1*A-k2*B)` if I am not mistaken.

- So the forward law would be `compartment*k1*A`.
- And the reverse law would be `compartment*k2*B`.

In MTK, we would translate the `Reaction` to
```Julia
Reaction(compartment*k1*A, [A], [B], [1], [1], use_only_rates=true)
Reaction(compartment*k2*B, [B], [A], [1], [1], use_only_rates=true)
```
In `Catalyst` I could express this as a single reversible reaction, with a 2-Tuple of `kineticLaws`. So there is no way around splitting up the `kineticLaws` (is this correct @isaacsas?).

isaacsas:
There is no way to avoid splitting them up currently if you go through a `ReactionSystem`; we fundamentally only store `Reaction`s, so the base information has to be at that level. 

As a future improvement, not something to worry about for the first version, it would be much better to detect if the rate law is mass action and reduce to
```julia
Reaction(compartment*k1, [A], [B], [1], [1])
Reaction(compartment*k2, [B], [A], [1], [1])
```
While this won't make a difference for ODE models, for jump models performance is noticeably better when using `MassActionJump`s to represent a `Reaction`, which is only possible when `use_only_rates=false`. Using `use_only_rates=true` may have other implications that can lead to slower system construction / conversion on large models (if used a lot). A `Reaction` built with `use_only_rates=true` is going to give `ismassaction(rx)==false` too I think, which isn't really right in this case.

paulflang:
So we would need to

- have a function that tells us whether a kineticLaw is mass action given the reactants and their stoichiometry.
- have a function that removes all the Species from the kineticLaw if it is mass action (e.g. `k1*A*B` becomes just `k1`).

Do MTK/Catalyst already contain such functions?

isaacsas:
We could add an API function that given the reactant and product stoichiometries, and the reactant and product species, checks if the corresponding mass action rate law equals the rate law expression from SBML (as a symbolics.jl expression). The only caveat is this would come down to how well symbolics can check equality of the two expressions, but my guess is that most cases are pretty simple and so would hopefully agree.

isaacsas:
Actually, it is probably better to look at the ratio of the two rate laws, using a rate constant of one for the mass action rate law, and check it is independent of any species variables? Assuming symbolics can simplify that appropriately the remainder then gives us the rate constant too. For your example we'd have
```julia
compartment*k1*A / A = compartment*k1
```
if simplification works, pulling out the exact rate expression to use with Catalyst.

paulflang:
Fantastic! That seems like the way to go.

exaexa:
> ```julia
> compartment*k1*A / A = compartment*k1
> ```

The domains of LHS/RHS are not equal so this can't check by default, you'd probably need to specify that `A!=0` can be assumed.

laurentheirendt (approved):

anandijain:
🎉  thanks for your help all!

PR #40, opened 2021-04-14
draft `Model` to MTK.___System conversion.

paulflang:
`reactionsystem.jl` should probably eventially be moved to the `ModelingToolkit.jl` package (to get rid of all the ModelingToolkit/Catalyst/Symbolics dependencies here). Other things to clarify:

- Model needs to have localParameters promoted.
- We need to figure out how to make `kineticLaws` extensive (see #31 )
- We need to figure out how to extract forward and reverse `kineticLaws` from Reactions where `reversible=true`.
- We need to decide on the type of `Reaction.kinetic_math`. This will allow to convert it to the right MTK symbols.


src/reactionsystem.jl
```diff
@@ -0,0 +1,127 @@
...
+function ModelingToolkit.ODESystem(model::Model)
+    rs = ReactionSystem(model)
+    u0map = get_u0(model)
+    parammap = get_paramap(model)
+    defaults = vcat(u0map, parammap)
+    convert(ODESystem, rs, defaults=defaults)
+end
+
+""" ODESystem constructor """
+function ModelingToolkit.ODESystem(sbmlfile::String)
```
anandijain, line 37: eventually need to fix the kwargs, so that `@named mymodel = ODESystem(sbmlfile)` will work

paulflang: where can I find documentation of the `@named` macro. I've never used it.

anandijain:
```julia 
help?> @named
  Rewrite @named y = foo(x) to y = foo(x; name=:y).
```

anandijain: I think you basically just need to change this to 
`function ModelingToolkit.ODESystem(sbmlfile::String; kwargs...)`
and then pass them to the actual constructor

src/reactionsystem.jl
```diff
@@ -0,0 +1,127 @@
...
+
+""" Convert intensive to extensive expressions """
+function make_extensive(model)
+    model  # Todo: For spevies with `hOSU=false` divide all occurences in mathematical expressions by compartment size.
+           # Also convert species initialConcentrations to initialAmounts
+end
+
+""" Expand reversible reactions to two reactions """
+function expand_reversible(model)
+    model  # Todo: convert all Reactions that are `reversible=true` to a forward and reverse reaction with `reversible=false`.
```
anandijain, line 62: Do we need to do this if `X + Y ↔ XY` is valid catalyst? https://catalyst.sciml.ai/stable/tutorials/basics/#Combining-several-reactions-in-one-line

Alternatively, it could be something that Catalyst needs. It seems like a useful transform to have.

anandijain: Maybe worth pinging Sam about

paulflang: The way as SBML would describe the kineticLaw for a reversible reaction is `X + Y ↔ XY` is `kAs*X*Y-kDi*XY`. Catalyst however needs `kAs*X*Y, kDi*XY`, that is two kineticLaws. I mean this is a simple example, but somehow we need to find a way that generally works in splitting a combined kineticLaw in a forward and reverse kineticLaw. Any suggestions how to do this in a clean way?

src/reactionsystem.jl
paulflang: Thanks!

exaexa:
So I finally got to reading this, quick comments:

> reactionsystem.jl should probably eventially be moved to the ModelingToolkit.jl package (to get rid of all the ModelingToolkit/Catalyst/Symbolics dependencies here).

Yeah, I'm not sure that others would really welcome pulling in the extra dependencies. Also, the code is very ODE specific.

> We need to figure out how to extract forward and reverse kineticLaws from Reactions where reversible=true.

Does this mean that there are cases that reaction has 2 different math rules that get triggered by changing the reaction direction?

> We need to decide on the type of Reaction.kinetic_math. This will allow to convert it to the right MTK symbols.

I'll keep it at `SBML.Math`, it is the best representation of how libsbml "parses" the MathML. Also, writing a translator from it to any other representation is super easy, I can probably write it. Can you send me an example of the MTK symbols you want to have at the end of the conversion?

paulflang:
> Yeah, I'm not sure that others would really welcome pulling in the extra dependencies. Also, the code is very ODE specific.

Yeah, I just thought for developing purposes it is easier if it is in the same repo. But can be moved now or whenever we feel like things have stabilised. What do you think?

> Does this mean that there are cases that reaction has 2 different math rules that get triggered by changing the reaction direction?

Not quite, if I understand you correctly. It just means that that you have a single math rule for the reaction rate which can be positive or negative, depending on whether the forward or reverse reaction is dominating. Consider the reaction `X + Y ↔ XY` where SBML would set `reversible=true`. This reaction would not have seperate `kineticLaw` for forward and reverse reaction. Rather it would have a single rate law `kAs*X*Y-kDi*XY`. However, for conversion into a `ReactionSystem` we need `kAs*X*Y` and `kDi*XY` as two seperate `kineticLaws`.

> Also, writing a translator from it to any other representation is super easy, I can probably write it. Can you send me an example of the MTK symbols you want to have at the end of the conversion?

I think SBML.jl should be agnostic of MTK (in the long run). It would make more sense to convert to [Symbolics.jl](https://github.com/JuliaSymbolics/Symbolics.jl) symbols, I believe.
Long story short, everything that is a `<ci>x</ci>` should become `Num(Variable(Symbol(x)))`, `<times/>` should become the Julia `*` operator, `plus/>` the Julia `+` operator, etc. And then it should be straightforward for me to convert that to the corresponding MTK types.
But @anandijain knows more about that.

paulflang:
For an XML example of bidirectional reaction you can look at #39

anandijain:
Yeah to that very last point. You can quickly see how MathML does this lowering here https://github.com/SciML/MathML.jl/blob/main/src/maps.jl. Basically I just string match the name of the first element of the "apply" XML node and return the equivalent Julia function

anandijain:
Why was this closed?

laurentheirendt:
it was probably automatically closed when I merged #39 ...

paulflang:
@fbergmann or @skeating: When an SBML reaction is reversible, it still only contains a single `kineticLaw`. Is there a function in libSBML that separates the single `kineticLaw` into a forward and reverse `kineticLaw`?

src/reactionsystem.jl
```diff
@@ -0,0 +1,152 @@
...
+        elseif v > 0
+            push!(products, create_var(k))
+            push!(pstoich, -v)
+        else
+            @error("Stoichiometry of $k must be non-zero")
+        end
+    end
+    subsdict = _get_substitutions(model)
+    # PL: Todo: @Anand: can you convert kinetic_math to Symbolic expression. Perhaps it would actually better if kinetic Math would be a Symbolics.jl expression rather than of type `Math`? But Mirek wants `Math`, I think.
+    kl = substitute(reaction.kinetic_math, subsdict)  # PL: Todo: might need conversion of kinetic_math to Symbolic MTK expression
```
anandijain, line 109: IIRC @exaexa was taking this Math* sublanguage conversion. But I can get started on it for sure

Manifest.toml
```diff
@@ -1,5 +1,34 @@
 # This file is machine-generated - editing it directly is not advised
 
+[[AbstractAlgebra]]
```
anandijain, line 3: delete this file

src/reactionsystem.jl
```diff
@@ -0,0 +1,165 @@
...
+    # model = expand_reversible(model)
+    rxs = mtk_reactions(model)
+    t = Catalyst.DEFAULT_IV
+    species = [create_var(k) for k in keys(model.species)]
+    params = vcat([create_param(k) for k in keys(model.parameters)], [create_param(k) for k in keys(model.compartments)])
+    ReactionSystem(rxs,t,species,params; kwargs...)
+end
+
+""" ReactionSystem constructor """
+function ModelingToolkit.ReactionSystem(sbmlfile::String; kwargs...)
```
anandijain, line 22: I'm not sure we will want to steal the String dispatch, since there are other formats that create a ReactionSystem (BioNetGen IIRC). 

I think just having `ModelingToolkit.ReactionSystem(model::Model; kwargs...)` might be good enough.
It's two function calls, but saves us from committing type piracy down the line.

src/reactionsystem.jl
```diff
@@ -0,0 +1,165 @@
...
+    rs = ReactionSystem(model; kwargs...)
+    model = make_extensive(model)  # PL: consider making `make_extensive!` to avoid duplicate calling in ReactionSystem and here
+    u0map = get_u0(model)
+    parammap = get_paramap(model)
+    defaults = Dict(vcat(u0map, parammap))
+    convert(ODESystem, rs, defaults=defaults)
+end
+
+""" ODESystem constructor """
+function ModelingToolkit.ODESystem(sbmlfile::String; kwargs...)
```
anandijain, line 38: same for this dispatch, I think this is type-piracy https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy. It's not a big deal ATM though, so I'm fine with it

paulflang: Not a problem now, I think:
```
julia> using ModelingToolkit, Catalyst

julia> methods(ReactionSystem)
# 3 methods for type constructor:
[1] ReactionSystem(iv; kwargs...) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\reaction\reactionsystem.jl:161
[2] ReactionSystem(eqs, iv, species, params; observed, systems, name) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\reaction\reactionsystem.jl:152
[3] ReactionSystem(eqs, iv, states, ps, observed, name, systems) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\reaction\reactionsystem.jl:147

julia> methods(ODESystem)
# 6 methods for type constructor:
[1] ODESystem(eq::Equation, args...; kwargs...) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\diffeqs\odesystem.jl:226
[2] ODESystem(eqs::Vector{Equation}, iv::Sym, states::Vector{T} where T, ps::Vector{T} where T, observed::Vector{Equation}, tgrad::Base.RefValue{Vector{Num}}, jac::Base.RefValue{Any}, Wfact::Base.RefValue{Matrix{Num}}, 
Wfact_t::Base.RefValue{Matrix{Num}}, name::Symbol, systems::Vector{ODESystem}, defaults::Dict, structure, reduced_states::Vector{T} where T) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\diffeqs\odesystem.jl:26
[3] ODESystem(deqs::AbstractVector{var"#s6"} where var"#s6"<:Equation, iv, dvs, ps; observed, systems, name, default_u0, default_p, defaults) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\diffeqs\odesystem.jl:75
[4] ODESystem(eqs) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\diffeqs\odesystem.jl:137
[5] ODESystem(eqs, iv; kwargs...) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\diffeqs\odesystem.jl:137
[6] ODESystem(eqs, iv, states, ps, observed, tgrad, jac, Wfact, Wfact_t, name, systems, defaults, structure, reduced_states) in ModelingToolkit at C:\Users\wolf5212\.julia\packages\ModelingToolkit\Bb2DA\src\systems\diffeqs\odesystem.jl:26
```

paulflang: The real question is where reaction network importers like our SBML importer (but also CellML and BioNetGen importers) should go in the end. I am not yet convinced that having separate packages for every model specification language is the right way to go. What type of file is handled can be automatically inferred from the ending or header given a filename as String.
We could put the importers in:
- ModelingToolkit.jl: For instance, we could dispatch the filename as `String` to the `ReactionSystem` constructor: `ModelingToolkit.ReactionNetwork(filename::String)`
- ReactionNetworkImporters.jl: For instance, we could dispatch a `NetworkFileFormat` constructor such as `BNGNetwork()` and filename to `loadrxnnetwork`: `ReactionNetworkImporters.loadrxnetwork(BNGNetwork(), filename)`

@isaacsas: any thoughts?

isaacsas:
If anything, `ReactionSystem`s will move from ModelingToolkit to Catalyst at some point, so I'd say this stuff should either go in Catalyst or in ReactionNetworkImporters. @ChrisRackauckas are there more breaking changes in line for core Symbolics/ModelingToolkit functionality, or is now a good time to move `ReactionSystem`s to Catalyst? 

If we want a single file-loading library, it would be fine with me to move ReactionNetworkImporters to SciML and host everything in it. Not everyone will want to use the file-loadling functionality, and it may ultimately become a heavier dependency as the library grows, so having it as a stand-alone package makes sense. I'm in no way wedded to the design in ReactionNetworkImporters, so we can change around the interface in a way that works best across multiple file formats too.

ChrisRackauckas:
>If anything, ReactionSystems will move from ModelingToolkit to Catalyst at some point, so I'd say this stuff should either go in Catalyst or in ReactionNetworkImporters. @ChrisRackauckas are there more breaking changes in line for core Symbolics/ModelingToolkit functionality, or is now a good time to move ReactionSystems to Catalyst?

Now might be a good time? I don't know how array symbolics will effect things here, but I would assume most of the changes would be downstream of ReactionSystem. Things have mostly stabilized by now.

I'm fine with however we structure the importers. One library or 3. Either way, I think ReactionNetworkImporters in SciML would be good because I always think distributing the maintenance and access is always better.

paulflang:
I think we are getting closer. Conversion to Symbolics and handling of hOSU is now implemented. But some minor test cases are failing. In particular could you have a look why this is failing, @anandijain :
```
Model to MTK conversions: Test Failed at C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\test\reactionsystem.jl:62
  Expression: isequal(Catalyst.get_states(odesys), [s1, s2, s1s2])
   Evaluated: isequal(Sym{Real, Nothing}[s1, s1s2, s2], Num[s1, s2, s1s2])
```
It could have to do with line 172 in https://github.com/paulflang/SBML.jl/blob/pl/mk-species-units/src/reactionsystem.jl

anandijain:
Maybe see if Set(get_states)(...)) is equal, since it looks out of order.

paulflang:
Thanks! And I thought I've got the types wrong.

src/reactionsystem.jl
paulflang: @shahriariravanian: what are your plans for CellMLToolkit. Do you think it should be moved under ReactionNetworkImporters.jl (see also Sam's and Chris' comment below)? Or shall it remain a standalone library?

shahriariravanian: I think eventually it is a good idea to put all of them under the same package, but ReactionNetworkImporters doesn't seem like the right place for CellMLToolkit. 

First, CellMLToolkit mainly outputs ODESystems (and possibly DAESystems in the next version), not reaction networks. While CellMLToolkit specification can handle reaction networks, very few actual CellML files contain one (the same that SBML files can define an ODE, but most don't). For whatever historical reason, CellML and SBML have diverged to their own niches. 

Second, while it is currently only an importer, the future plan for CellMLToolkit is -- after adding DAEs and implementing units handling and Unitful integration -- to write an exporter. I don't know if you plan to add export functionality to SBML, but I feel that to use the full benefits of symbolic manipulation, we need to have the capability to output the results.

anandijain:
So is this ready for review and undrafted? @paulflang 

Also again amazing job 👏👏👏👏👏

exaexa:
Morning! Let's convert this to a non-draft and see. :]

src/readsbml.jl
```diff
@@ -96,6 +96,39 @@ function readSBML(fn::String)::SBML.Model
             throw(AssertionError("SBML document contains no model"))
         end
 
+        for (converter, kwargs) in conversion_options
```
exaexa, line 99: extending the conversions this way requires you to extend SBML (thus basically making it non-extensible from the user side). You can pass in a normal conversion function to make this much more maintainable. (OOP folks would call it "callback" or "functor" or something)

src/readsbml.jl
```diff
@@ -96,6 +96,39 @@ function readSBML(fn::String)::SBML.Model
...
+            end
+        end
+
+        n_errs = ccall(sbml(:SBMLDocument_getNumErrors), Cuint, (VPtr,), doc)
+        for i = 0:n_errs-1
+            err = ccall(sbml(:SBMLDocument_getError), VPtr, (VPtr, Cuint), doc, i)
+            msg = SBML.get_string(err, :XMLError_getMessage)
+            @warn "SBML reported error after conversion: $msg"
+        end
+        if n_errs > 0  # PL: @Mirek: Maybe we should only throw an error if certain severity is exceeded using `SBMLDocument_getNumErrorsWithSeverity`?
```
exaexa, line 123: I hoped that this excludes the "info" and "warning" severities.

(SBML error severities are more like logleves, they don't carry any semantics that would tell you whether it's feasible to continue, e.g. if a data loss was likely. So we likely can't solve this one well.)

exaexa: btw given this is going to be externalized to a function, you can very easily leave that decision on the user.

src/readsbml.jl
```diff
@@ -228,8 +261,16 @@ function extractModel(mdl::VPtr)::SBML.Model
...
+        if ccall(sbml(:Species_isSetInitialConcentration), Cint, (VPtr,), sp) != 0
+            ic = (
+                ccall(sbml(:Species_getInitialConcentration), Cdouble, (VPtr,), sp),
+                get_string(sp, :Species_getSubstanceUnits),
+            )
+        end
+
         species[get_string(sp, :Species_getId)] = Species(
-            get_optional_string(sp, :Species_getName),
+            get_optional_string(sp, :Species_getName),  # PL: wondering if we should use Id instead of name here
```
exaexa, line 273: identifiers are parsed as identifiers (ie keys), that cannot be confused with (completely optional and computtionally irrelevant) "pretty name".

src/sbml2symbolics.jl
```diff
@@ -0,0 +1,31 @@
+using ModelingToolkit, Symbolics
+
+allowed_funs = Dict(
+    "+" => :+,
+    "-" => :-,
+    "*" => :*,
+    "/" => :/,
+    #TODO add translations of SBML functions to Julia
```
exaexa, line 8: TODO

src/sbml2symbolics.jl
```diff
@@ -0,0 +1,31 @@
...
+)
+
+allowed_sym(x) = haskey(allowed_funs,x) ? allowed_funs[x] : DomainError(x,"Unknown SBML function")
+
+function SBML2Symbolics(x::Math)
```
exaexa, line 13: there's the standard way of converting stuff in julia, I'd overload Base.convert

src/sbml2symbolics.jl
```diff
@@ -0,0 +1,31 @@
...
+end
+
+
+# SBML.MathApply("*", SBML.Math[
+#     SBML.MathApply("*", SBML.Math[
+#         SBML.MathApply("*", SBML.Math[
+#             SBML.MathIdent("c1"),
+#             SBML.MathIdent("k1")]),
+#         SBML.MathIdent("s1")]),
+#     SBML.MathIdent("s2")])
```
exaexa, line 31: please clean up the comment notes. Also please check that all lines have \n on the end (so that the ugly label on the end of the last line disappears :] ). There's a switch for that in VSCode, but it's off by default for whatever bad reason.

src/structs.jl
```diff
@@ -77,7 +77,7 @@ end
 """
 Function application ("call by name", no tricks allowed) in mathematical expression
 """
-struct MathApply <: Math
+mutable struct MathApply <: Math
```
exaexa, line 80: please do not use mutable structures. The Model represents something that's in a file, which should not be mutable.

src/structs.jl
```diff
@@ -110,11 +110,11 @@ consumption/production rates (accessible in field `stoichiometry`), lower/upper
 bounds (in tuples `lb` and `ub`, with unit names), and objective coefficient
 (`oc`). Also may contains `notes` and `annotation`.
 """
-struct Reaction
-    stoichiometry::Dict{String,Float64}
-    lb::Tuple{Float64,String}
-    ub::Tuple{Float64,String}
-    oc::Float64
+mutable struct Reaction
```
exaexa, line 113: mutable

src/structs.jl
```diff
@@ -110,11 +110,11 @@ consumption/production rates (accessible in field `stoichiometry`), lower/upper
...
+    stoichiometry::Dict{String,Float64}  # PL: Why do we allow/require non-integer stoichiometries?
```
exaexa, line 114: Unfortunately it seems nature is not made of integers, at least not at this zoom level. :D

SBML allows full floats here and many models use that. (There's even functionality that uses additional denominators to write `0.33333333` precisely as `1/3`, but let's pretend that one doesn't exist.)

src/structs.jl
```diff
@@ -126,18 +126,19 @@ end
 Species metadata -- contains a human-readable `name`, a `compartment`
 identifier, `formula`, `charge`, and additional `notes` and `annotation`.
 """
-struct Species
+mutable struct Species
```
exaexa, line 129: mutable

src/structs.jl
```diff
@@ -175,8 +176,8 @@ struct Model
     compartments::Dict{String,Compartment}
     species::Dict{String,Species}
     reactions::Dict{String,Reaction}
-    gene_products::Dict{String,GeneProduct}
-    function_definitions::Dict{String,FunctionDefinition}
+    gene_products::Maybe{Dict{String,GeneProduct}}
+    function_definitions::Maybe{Dict{String,FunctionDefinition}}
```
exaexa, line 180: Why have Maybe here? IMHO it may make some sense for the gene products (there might be a difference between "there are no gene products linked" vs "we don't talk about gene products here"), but certainly not for fundefs, if these are empty then they are just empty with no additional semantics right?

test/loaddynamicmodels.jl
```diff
@@ -0,0 +1,55 @@
+using ModelingToolkit
+
+@parameters t, k1, c1
```
exaexa, line 3: move to testset scope

test/loaddynamicmodels.jl
```diff
@@ -0,0 +1,55 @@
+using ModelingToolkit
```
exaexa, line 1: load centrally in runtests.jl

test/loaddynamicmodels.jl
```diff
@@ -0,0 +1,55 @@
...
+
+@parameters t, k1, c1
+@variables s1, s2, s4
+
+sbmlfiles = [
+    # a test w/o conversion
```
exaexa, line 7: ?

test/loaddynamicmodels.jl
```diff
@@ -0,0 +1,55 @@
...
+    for (sbmlfile, conversion_options, expected_pars, expected_rxn) in sbmlfiles
+        @testset "Loading of $sbmlfile" begin
+            mdl = readSBML(sbmlfile;conversion_options=conversion_options)
+
+            @test typeof(mdl) == Model
+
+            @test length(mdl.parameters) == expected_pars
+
+            id, math = expected_rxn
+            # @test mdl.reactions[id].kinetic_math == math  # PL: Todo: Comment in once MathML is done
```
exaexa, line 50: todo

test/reactionsystem.jl
```diff
@@ -0,0 +1,132 @@
+using Catalyst, ModelingToolkit, OrdinaryDiffEq
```
exaexa, line 1: centralize imports to runtests.jl

test/reactionsystem.jl
```diff
@@ -0,0 +1,132 @@
...
+COMP1 = SBML.Compartment("c1", true, 3, 2., "nl") 
+SPECIES1 = SBML.Species("s1", "c1", false, nothing, nothing, (1., "substance"), nothing, true)  # Todo: Maybe not support units in initial_concentration?
+SPECIES2 = SBML.Species("s2", "c1", false, nothing, nothing, nothing, (1., "substance/nl"), false)
+KINETICMATH1 = SBML.MathIdent("k1")
+KINETICMATH2 = SBML.MathApply("*", SBML.Math[
+    SBML.MathIdent("k1"), SBML.MathIdent("s2")])
+REACTION1 = SBML.Reaction(Dict("s1" => 1), nothing, nothing, nothing, nothing, KINETICMATH1)
+REACTION2 = SBML.Reaction(Dict("s2" => -1), nothing, nothing, nothing, nothing, KINETICMATH2)
+MODEL1 = SBML.Model(Dict("k1" => 1.), Dict(), Dict("c1" => COMP1), Dict("s1" => SPECIES1), Dict("r1" => REACTION1), nothing, nothing)  # PL: For instance in the compartments dict, we may want to enforce that key and compartment.name are identical
+MODEL2 = SBML.Model(Dict("k1" => 1.), Dict(), Dict("c1" => COMP1), Dict("s2" => SPECIES2), Dict("r2" => REACTION2), nothing, nothing)
```
exaexa, line 21: move everything to testset scope
(btw why SCREAM_CAPS?)

test/reactionsystem.jl
```diff
@@ -0,0 +1,132 @@
...
+    model = SBML.to_initial_amounts(MODEL2)
+    @test isequal(model.species["s2"].initial_amount, (2., ""))
+    @test isequal(model.species["s2"].initial_concentration, nothing)
+
+    # Test to_extensive_math!
+    model = SBML.to_extensive_math!(MODEL1)
+    @test isequal(model.reactions["r1"].kinetic_math, KINETICMATH1)
+    model = SBML.to_extensive_math!(MODEL2)
+    @test isequal(repr(model.reactions["r2"].kinetic_math), repr(kineticmath2_true))
+    #PL: Todo @Mirek/Anand: Why does this not work without the `repr()`?
```
exaexa, line 106: what precise part of the structure fails the comparison?

test/reactionsystem_01.xml
```diff
@@ -0,0 +1,162 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
exaexa, line 1: save the test files in `test/data/` so that they don't mix up with code.

(That reminds me I should properly .gitignore the downloaded files.)

test/runtests.jl
```diff
@@ -5,7 +5,10 @@ using SBML: Model, Reaction, Species
 import Pkg
 
 @testset "SBML test suite" begin
-    include("version.jl")
-    include("ecoli_flux.jl")
-    include("loadmodels.jl")
+    # include("version.jl")
```
exaexa, line 8: ?

test/sbml2symbolics.jl
```diff
@@ -0,0 +1,43 @@
+using Symbolics
+
+# sbmlfile = "reactionsystem_01.xml"
```
exaexa, line 3: ?

test/sbml2symbolics.jl
```diff
@@ -0,0 +1,43 @@
...
+# COMP1 = Compartment("c1", true, 3, 2., "nl") 
+# SPECIES1 = Species("s1", "c1", false, nothing, nothing, (1., "substance"), nothing, true)  # Todo: Maybe not support units in initial_concentration?
+# SPECIES2 = Species("s2", "c1", false, nothing, nothing, nothing, (1., "substance/nl"), false)
+# KINETICMATH1 = SBML.MathIdent("k1")  # PL: @anand or @mirek help needed. Can someone create `k1 * SPECIES1` as a `Math` type here?
+# REACTION1 = SBML.Reaction(Dict("s1" => -1), nothing, nothing, nothing, nothing, KINETICMATH1)
+# REACTION2 = SBML.Reaction(Dict("s2" => -1), nothing, nothing, nothing, nothing, KINETICMATH1)
+# MODEL1 = Model(Dict("k1" => 1.), Dict(), Dict("c1" => COMP1), Dict("s1" => SPECIES1), Dict("r1" => REACTION1), nothing, nothing)  # PL: For instance in the compartments dict, we may want to enforce that key and compartment.name are identical
+# MODEL2 = Model(Dict("k1" => 1.), Dict(), Dict("c1" => COMP1), Dict("s2" => SPECIES2), Dict("r2" => REACTION2), nothing, nothing)
+
+A = Symbolics.Variable(Symbol("A"))
```
exaexa, line 21: move to testset scope

test/reactionsystem.jl
exaexa: btw why `isequal`? The normal `==` seems more appropriate here to me.

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
...
+        push!(paramap,Pair(create_param(k),v.size))
+    end
+    paramap
+end
+
+create_var(x) = Num(Variable(Symbol(x))).val
+function create_param(x)
+    p = Sym{Real}(Symbol(x))
+    ModelingToolkit.toparam(p)
+end
```
exaexa, line 176: missing endline

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
...
+    ODESystem(model; kwargs...)
+end
+
+""" ODEProblem constructor """
+function ModelingToolkit.ODEProblem(model::Model,tspan;kwargs...)  # PL: Todo: add u0 and parameters argument
+    odesys = ODESystem(model)
+    ODEProblem(odesys, [], tspan; kwargs...)
+end
+
+""" ODEProblem constructor """
```
exaexa, line 42: regarding the docstrings, please include _full_ function types and signatures, and preferably some hint about what precisely the function does it & how.

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
+using Catalyst, ModelingToolkit, Symbolics
```
exaexa, line 1: centralize imports to src/SBML.jl

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
...
+    #         throw(DomainError(reaction, "Cannot convert `kineticLaw` of reversible reaction to MTK.Reaction.rate"))
+    #     end
+    # end
+    return  # Todo: convert all Reactions that are `reversible=true` to a forward and reverse reaction with `reversible=false`.
+end
+
+""" Convert intensive to extensive expressions """
+function make_extensive(model)
+    model = to_initial_amounts(model)
+    model = to_extensive_math!(model)
```
exaexa, line 61: is this an inplace modification (implied by !) or a normal new-value-returning function?

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
...
+end
+
+""" Check if conversion to ReactionSystem is possible """
+function checksupport(model)
+    # for reaction in model.reactions
+    #     if reaction.reversible
+    #         throw(DomainError(reaction, "Cannot convert `kineticLaw` of reversible reaction to MTK.Reaction.rate"))
+    #     end
+    # end
+    return  # Todo: convert all Reactions that are `reversible=true` to a forward and reverse reaction with `reversible=false`.
```
exaexa, line 55: todo

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
...
+    model = to_initial_amounts(model)
+    model = to_extensive_math!(model)
+    model
+end
+
+""" Convert initial_concentration to initial_amount """
+function to_initial_amounts(model::Model)
+    model = deepcopy(model)
+    for specie in values(model.species)
+        if isequal(specie.initial_amount, nothing)
```
exaexa, line 69: isnothing() ?

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
...
+""" Convert intensive to extensive expressions """
+function make_extensive(model)
+    model = to_initial_amounts(model)
+    model = to_extensive_math!(model)
+    model
+end
+
+""" Convert initial_concentration to initial_amount """
+function to_initial_amounts(model::Model)
+    model = deepcopy(model)
```
exaexa, line 67: you don't need a full copy here, just copy()ing the model struct and deepcopying the species is okay.

src/reactionsystem.jl
```diff
@@ -0,0 +1,176 @@
...
+            compartment = model.compartments[specie.compartment]
+            specie.initial_amount = (specie.initial_concentration[1] * compartment.size, "")
+            specie.initial_concentration = nothing
+        end
+    end
+    model
+end
+
+""" Convert intensive to extensive mathematical expression """
+function to_extensive_math!(model::SBML.Model)
```
exaexa, line 79: I find the intensive vs. extensive naming a bit confusing, but that's maybe because I didn't work with this much.

If I get it correctly, this here means basically "substitute the compartment sizes" right?

exaexa (changes requested):

test/reactionsystem.jl
paulflang: In Python it seems to be convention that constants in the global scope are in SCREAM_CAPS. I am using these here throughout the test cases. What is the convention in Julia?

ChrisRackauckas: same convention.

test/reactionsystem.jl
paulflang: > what precise part of the structure fails the comparison?

```
Model to MTK conversions: Test Failed at C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\test\reactionsystem.jl:90
  Expression: isequal((model.reactions["r2"]).kinetic_math, kineticmath2_true)
   Evaluated: isequal(SBML.MathApply("*", SBML.Math[SBML.MathIdent("k1"), SBML.MathApply("*", SBML.Math[SBML.MathVal{Float64}(2.0), SBML.MathIdent("s2")])]), SBML.MathApply("*", SBML.Math[SBML.MathIdent("k1"), SBML.MathApply("*", SBML.Math[SBML.MathVal{Float64}(2.0), SBML.MathIdent("s2")])]))
```
   
> btw why isequal? The normal == seems more appropriate here to me.

Idk, @anandijain and I have been using it like that. @anandijain did you have a reason?

src/reactionsystem.jl
paulflang: @ChrisRackauckas @isaacsas @shahriariravanian @anandijain : After @shahriariravanian 's response that CellMLToolkit will grow bigger and does not fit well under the ReactionNetworkImporters umbrella I am inclined to have all Toolkits for biomodel importers as independent packages. Most users have their favourite model specification language and will ever only need one importer, so there are few reasons to combine them in one package. The only reason I can think of, is to make sure they have a similar API. But for the few users who use two or more model specification languages we can still take care that the API looks similar across importers.

To this end, I could imagine mimicking the constructors of the `MTK.AbstractSystem` types without overloading them to avoid type piracy. A bit like so (or is this a bad idea?):

```
module MTK
ODEsys(x) = println(x)
export ODEsys
end

module CellMLTK
using Main.MTK
ODEsys(x::String) = Main.MTK.ODEsys("Creating ODEsys from CellML file")
end

module SbmlTK
using Main.MTK
ODEsys(x::String) = Main.MTK.ODEsys("Creating ODEsys from SBML file")
end

using Main.MTK
x = "Creating ODEsys from scratch"
ODEsys(x)  # Creating ODEsys from scratch

x="Cannot create ODEsys from filename"
ODEsys(x)  # Cannot create ODEsys from filename
Main.CellMLTK.ODEsys(x)  # Creating ODEsys from CellML file
Main.SbmlTK.ODEsys(x)  # Creating ODEsys from SBML file
```

ChrisRackauckas: Please don't shadow, just extend.

test/reactionsystem.jl
anandijain: Mk, isequal is the way to compare symbolics, i suggest just messing around w boolean operators in symbolics and it will make sense

src/reactionsystem.jl
paulflang: addressed via ce4f03b3c1ae19c68f19

src/reactionsystem.jl
paulflang: fixed with 537296b74c3233b978809cc0da42b2794927b7a1

src/reactionsystem.jl
paulflang: Yes it is confusing. Even the SBML specs confused it (should be fixed now or soon). But it has a clear definition in physics and I cannot think of a better terminology.

src/reactionsystem.jl
paulflang: To avoid misunderstandings, I assume shadowing is creating another function with the same name that takes precedence over the old function. And extending means overloading with a more specific type to make use of multiple dispatch. Is this correct?

Anyway, I do not get it to work. Especially not without avoiding type piracy.
```
module MTK
ODEsys(x) = println(x)
export ODEsys
end
module CellMLTK
using Main.MTK
Main.MTK.ODEsys(x::String) = ODEsys("Creating ODEsys from CellML file")
end
# module SbmlTK
# using Main.MTK
# Main.MTK.ODEsys(x::String) = ODEsys("Creating ODEsys from SBML file")
# export ODEsys
# end
using Main.MTK, Main.CellMLTK  #, Main.SbmlTK
x = "Creating ODEsys from scratch"
ODEsys(x)  # Creating ODEsys from scratch
```
```
ERROR: StackOverflowError:
Stacktrace:
 [1] ODEsys(x::String) (repeats 79984 times)
   @ Main.CellMLTK .\REPL[2]:3
```
Do you have any suggestions?

src/reactionsystem.jl
paulflang: fixed with 5455600

src/readsbml.jl
paulflang: I've never worked with callbacks. Are you thinking of sth like this?

```Julia
# File readsbml.jl
function readSBML(fn::String;libsbml_converter::Maybe{Function}=nothing)::SBML.Model
    doc = ccall(sbml(:readSBML), VPtr, (Cstring,), fn)
    try
        ...

        # @Mirek: here happens the new conversion
        if !isnothing(libsbml_converter)
            libsbml_converter()
        end

        ...

        return extractModel(model)
    finally
        ccall(sbml(:SBMLDocument_free), Nothing, (VPtr,), doc)
    end
end


# File reactionsystem.jl
function libsbml_converter(conversion_options::Dict)

    for (converter, kwargs) in conversion_options
        if converter == "setLevelAndVersion"
            success = ccall(sbml(:SBMLDocument_setLevelAndVersion), Cint, (VPtr,Cint,Cint), doc, kwargs["level"], kwargs["version"])
            continue
        end
        props = ccall(sbml(:ConversionProperties_create), VPtr, ())
        option = ccall(sbml(:ConversionOption_create), VPtr, (Cstring,), converter)
        ccall(sbml(:ConversionProperties_addOption), Cvoid, (VPtr, VPtr), props, option)
        if !(kwargs==nothing)
            for (k, v) in kwargs
                option = ccall(sbml(:ConversionOption_create), VPtr, (Cstring,), k)
                ccall(sbml(:ConversionOption_setValue), Cvoid, (VPtr, Cstring), option, v)
                ccall(sbml(:ConversionProperties_addOption), Cvoid, (VPtr, VPtr), props, option)
            end
        success = ccall(sbml(:SBMLDocument_convert), Cint, (VPtr,VPtr), doc, props)
        end
    end
end

mtk_default_converter = function ()
    mtk_default_conversion_options = Dict("promoteLocalParameters" => nothing,
                                          "expandFunctionDefinitions" => nothing,
                                          "expandInitialAssignments" => nothing,
                                          "setLevelAndVersion" => Dict("level" => 3,
                                                                       "version" => 1)
                                         )
    libsbml_converter(mtk_default_conversion_options)
end

""" ReactionSystem constructor """
function ModelingToolkit.ReactionSystem(sbmlfile::String; libsbml_converter=mtk_default_converter, kwargs...)
    model = readSBML(sbmlfile;libsbml_converter=mtk_default_converter)
    ReactionSystem(model; kwargs...)
end
```

src/readsbml.jl
paulflang: I do not know what severities are excluded, but if I use the converters (for instance the `setLevelAndVersionConverter`) the AssertionError is thown because of the following errors, which I believe are not severe enough to stop the import:
```
┌ Warning: SBML reported error after conversion: In situations where a mathematical expression contains literal numbers or parameters whose units have not been declared, it is not possible to verify accurately the consistency of the units in the expression.
│  The units of the <kineticLaw> <math> expression 'c1 * k1 * s1 * s2' cannot be fully checked. Unit consistency reported as either no errors or further unit errors related to this object may not be accurate.
└ @ SBML C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\src\readsbml.jl:121
┌ Warning: SBML reported error after conversion: In situations where a mathematical expression contains literal numbers or parameters whose units have not been declared, it is not possible to verify accurately the consistency of the units in the expression.
│  The units of the <kineticLaw> <math> expression 'c1 * k1 * s3' cannot be fully checked. Unit consistency reported as either no errors or further unit errors related to this object may not be accurate.
└ @ SBML C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\src\readsbml.jl:121
┌ Warning: SBML reported error after conversion: In situations where a mathematical expression contains literal numbers or parameters whose units have not been declared, it is not possible to verify accurately the consistency of the units in the expression.
│  The units of the <kineticLaw> <math> expression 'c1 * function_01_1(s4)' cannot be fully checked. Unit consistency reported as either no errors or further unit errors related to this object may not be accurate.      
└ @ SBML C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\src\readsbml.jl:121
┌ Warning: SBML reported error after conversion: In SBML's simple symbol system, there is no separation of symbols by class of object; consequently, inside the kinetic law mathematical formula, the value of a local parameter having the same identifier as a species or compartment or other global model entity will override the global value. Modelers may wish to take precautions to avoid this happening accidentally.
│  In this instance the local parameter with id 'k1' will shadow the 'Parameter' with an identical id.
└ @ SBML C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\src\readsbml.jl:121
┌ Warning: SBML reported error after conversion: As a principle of best modeling practice, the units of a <parameter> should be declared rather than be left undefined. Doing so improves the ability of software to check 
the consistency of units and helps make it easier to detect potential errors in models.
│  The <parameter> with the id 'k1' does not have a 'units' attribute.
└ @ SBML C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\src\readsbml.jl:121
┌ Warning: SBML reported error after conversion: As a principle of best modeling practice, the units of a <parameter> should be declared rather than be left undefined. Doing so improves the ability of software to check 
the consistency of units and helps make it easier to detect potential errors in models.
│  The <localParameter> with the id 'k1' does not have a 'units' attribute.
└ @ SBML C:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\LCSB-BioCore\SBML.jl\src\readsbml.jl:121
```

paulflang: So can I get rid of throwing the `AssertionError` for the time being?

src/sbml2symbolics.jl
paulflang: I think this is best done when testing with the [sbml_test_suite](https://github.com/sbmlteam/sbml-test-suite/tree/master/cases/semantic). Then we see where the test cases fail and add translations on a case by case basis.

src/sbml2symbolics.jl
paulflang: fixed with 6913135

src/sbml2symbolics.jl
paulflang: done (I hope)

src/structs.jl
paulflang: Well, I need to change `Math` (and thus `Reaction`s) in `to_extensive_math!()`, and `Species` in `to_initial_amounts()`. It seems a bit cumbersome to write all this without mutating the `Math`, `Reaction`s and `Species`, but I can do it if needed. What do you think?

src/structs.jl
paulflang: In that case I think SBML allows stuff that nature doesn't (what on earth is a half-molecule?). But they sure have their reasons (e.g. some people might want to 1*H + 0.5*O -> 0.5*H2O). Thanks for the explanation, I was just puzzled.

src/structs.jl
paulflang: Makes sense. Fixed with a1af25a. I also took Maybe out of gene_products, because I do not see how the difference between "there are no gene products linked" vs "we don't talk about gene products here" can be seen in the SBML file.

test/loaddynamicmodels.jl
paulflang: fixed by 1ffe96a

test/loaddynamicmodels.jl
paulflang: fixed by 1ffe96a

test/loaddynamicmodels.jl
paulflang: fixed by 1ffe96a. Math conversion now happens downstream of SBML.Model, so this test is no longer relevant. But I remember we and @laurentheirendt were thinking about having Reaction.kinetic_math expressed in `Symbolics` rather than `SBML.Math` (see #47 ). Or am I mistaken?

test/loaddynamicmodels.jl
paulflang: fixed with dce53c3

test/reactionsystem.jl
paulflang: Thanks. I noticed that when testing Symbolics. Still, the question why the test fails if we do not do `repr` is not clear do me. It is certainly not an elegant way for testing.

test/reactionsystem.jl
paulflang: fixed with 44d88f5

test/reactionsystem.jl
paulflang: fixed with 44d88f5

src/structs.jl
paulflang: The changes I made in a1af25a caused test cases to break. fixed in 44d88f5.

test/reactionsystem_01.xml
paulflang: fixed with 69fa2ff

test/sbml2symbolics.jl
paulflang: I think this is fixed now, but I do not see the "Outdated" label.

test/sbml2symbolics.jl
paulflang: fixed with 3d0e09f

test/runtests.jl
paulflang: fixed with 667240f

src/reactionsystem.jl
```diff
@@ -0,0 +1,224 @@
...
+function ModelingToolkit.ODEProblem(model::Model,tspan;kwargs...)  # PL: Todo: add u0 and parameters argument
+    odesys = ODESystem(model)
+    ODEProblem(odesys, [], tspan; kwargs...)
+end
+
+""" ODEProblem constructor """
+function ModelingToolkit.ODEProblem(sbmlfile::String,tspan;kwargs...)  # PL: Todo: add u0 and parameters argument
+    odesys = ODESystem(sbmlfile)
+    ODEProblem(odesys, [], tspan; kwargs...)
+end
```
ChrisRackauckas, line 49: I would assume it would be best to have the user explicitly construct the ODESystem -> ODEProblem. This is one of those shorthands that causes doc problems.

src/readsbml.jl
paulflang: @exaexa : I tried to merge master into my my pl/mk-species-units branch and this topic becomes relevant again. Should I handle the conversion like that now?
```Julia
# File readsbml.jl
function readSBML(fn::String, sbml_conversion = document -> nothing)::SBML.Model
    doc = ccall(sbml(:readSBML), VPtr, (Cstring,), fn)
    try
        ...

        sbml_conversion(doc)

        ...

        return extractModel(model)
    finally
        ccall(sbml(:SBMLDocument_free), Nothing, (VPtr,), doc)
    end
end


# File reactionsystem.jl
function libsbml_converter(doc; conversion_options::Dict=Dict("promoteLocalParameters" => nothing, "expandFunctionDefinitions" => nothing, "expandInitialAssignments" => nothing, "setLevelAndVersion" => Dict("level" => 3, "version" => 1)))

    for (converter, kwargs) in conversion_options
        if converter == "setLevelAndVersion"
            success = ccall(sbml(:SBMLDocument_setLevelAndVersion), Cint, (VPtr,Cint,Cint), doc, kwargs["level"], kwargs["version"])
            continue
        end
        props = ccall(sbml(:ConversionProperties_create), VPtr, ())
        option = ccall(sbml(:ConversionOption_create), VPtr, (Cstring,), converter)
        ccall(sbml(:ConversionProperties_addOption), Cvoid, (VPtr, VPtr), props, option)
        if !(kwargs==nothing)
            for (k, v) in kwargs
                option = ccall(sbml(:ConversionOption_create), VPtr, (Cstring,), k)
                ccall(sbml(:ConversionOption_setValue), Cvoid, (VPtr, Cstring), option, v)
                ccall(sbml(:ConversionProperties_addOption), Cvoid, (VPtr, VPtr), props, option)
            end
        success = ccall(sbml(:SBMLDocument_convert), Cint, (VPtr,VPtr), doc, props)
        end
    end
end

""" ReactionSystem constructor """
function ModelingToolkit.ReactionSystem(sbmlfile::String; libsbml_converter=libsbml_converter, kwargs...)
    model = readSBML(sbmlfile;libsbml_converter=mtk_default_converter)
    ReactionSystem(model; kwargs...)
end
```

paulflang: Lets just create one converter:
```Julia
function libsbml_converter!(doc, converter::String; kwargs...)
    ...
    return doc
end

```
So you could then do
```Julia
readSBML("filename.xml", doc->begin
  libsbml_converter!(doc,"setLevelAndVersion"; level=3, version=1)
  libsbml_converter!(doc,"promoteLocalParameters")
end)
```

exaexa: That works now perfectly already:
```
readSBML("filename.xml", doc -> begin
  convert_level_and_version(3,1)(doc)
  libsbml_convert(Dict("promoteLocalParameters" => Dict{String,String}())
    )(doc)
end)
```
The (doc) parameters are curried to allow pointfree specification for single converters, but it works the same as before.
I could simplify the API of libsbml_convert though, the dictionary there looks more confusing then helpful.

src/structs.jl
paulflang: I think the to_extensive_math!() and to_initial_amounts() converters in reactionsystem.jl are generally useful beyond MTK and therefore should remain in SBML.jl. @exaexa : do you agree? Can you rewrite them so that they no longer need mutable structs? Or shall I give it a try (e.g. putting them into utils.jl)?

exaexa:
This is now partly merged and partly going to https://github.com/paulflang/SBMLToolkit.jl . Thanks!

PR #41, opened 2021-04-14
Delete Manifest.toml

anandijain:
Manifest.toml is in the .gitignore, but it is in the repo?


anandijain:
I'm not sure if this is sufficient (ie equivalent to `git rm --cached`)

laurentheirendt:
nice catch - thanks 👍🏼

exaexa:
just checking-- was this a remainder from the time we needed the CI to quickly take the newer version of SBML_jll?

laurentheirendt:
Yes ... 👍🏼

PR #49, opened 2021-05-28
unexports

anandijain:
not sure if you wanted to get rid of `Model` too, happy to revert that one

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -5,12 +5,12 @@ This package provides a straightforward way to load model- and
 simulation-relevant information from SBML files.
 
 The library provides a single function [`readSBML`](@ref) to load a
-[`Model`](@ref):
+[`SBML.Model`](@ref):
 
 ```julia
 julia> using SBML
 julia> mdl = readSBML("Ec_core_flux1.xml")
-Model(…)
+SBML.Model(…)
 
 julia> mdl.compartments
 2-element Array{String,1}:
--- a/src/SBML.jl
+++ b/src/SBML.jl
@@ -12,6 +12,6 @@ include("utils.jl")
 sbml = (sym::Symbol) -> dlsym(SBML_jll.libsbml_handle, sym)
 
 export SBMLVersion,
-    readSBML, UnitPart, getS, getLBs, getUBs, getOCs
+    readSBML, getS, getLBs, getUBs, getOCs
 
 end # module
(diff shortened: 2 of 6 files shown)


test/runtests.jl
```diff
@@ -1,6 +1,7 @@
 
 using Test, SHA, SparseArrays
 using SBML
+using SBML: Model, Reaction, Species
```
anandijain, line 4: I could also just change the test to say SBML.Model, but for now this is probably fine?

docs/src/index.md
```diff
@@ -1,81 +1,81 @@
-
```
exaexa, line 1: what happened to this file that the diff replaces _all_ lines?

src/SBML.jl
```diff
@@ -12,6 +12,6 @@ include("utils.jl")
 sbml = (sym::Symbol) -> dlsym(SBML_jll.libsbml_handle, sym)
 
 export SBMLVersion,
-    readSBML, Model, UnitPart, Species, Reaction, getS, getLBs, getUBs, getOCs
+    readSBML, UnitPart, getS, getLBs, getUBs, getOCs
```
exaexa, line 15: cool. Honestly I think UnitPart can get removed too.

test/runtests.jl
exaexa: yeah, this here is better.

src/utils.jl
```diff
@@ -1,54 +1,54 @@
...
-end
-
-"""
-    function getOCs(m::Model)::Vector{Float64}
-
-Extract the vector of objective coefficients of each reaction.
-"""
-function getOCs(m::Model)::Vector{Float64}
-    return broadcast(x -> x.oc, values(m.reactions))
-end
```
exaexa, line 54: also full replace instead of diff here?

exaexa (changes requested):

test/runtests.jl
anandijain: so this is resolved, right

docs/src/index.md
anandijain: yeah I don't know, I was concerned to see that too. 
Any ideas?

src/SBML.jl
anandijain: think i got this

src/utils.jl
anandijain: yes same issue, i simply did crtl f and replace all to change them.

Not sure how to prevent it

exaexa:
The code was transplanted to #50 to avoid CRLF/LF reformat, all merged now. Thanks a lot!

PR #55, opened 2021-06-07
add support for running SBML converters

exaexa:
This is the first part of disassembling #40 to tangible pieces.

Generalizes some of the conversion stuff also for other purposes. Conversion for MTK should be now doable easily with something like:
```julia
readSBML("animal.xml",
         doc -> begin
           convert_level_and_version(3, 1)(doc)
           convert_simplify_math(doc)
         end)
```

@paulflang is there some good way/data to test this?

### DIFF
--- a/src/SBML.jl
+++ b/src/SBML.jl
@@ -5,13 +5,15 @@ using SparseArrays
 
 include("structs.jl")
 include("version.jl")
+
 include("readsbml.jl")
+include("converters.jl")
 include("math.jl")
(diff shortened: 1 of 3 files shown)


exaexa:
Random midday complaint: if only we had monoids and applicatives in julia, the code would be so much neater
```jula
readSBML("animal.xml", convert_level_and_version(3,1) <> convert_simplify_math)
```

paulflang:

PR #63, opened 2021-06-15
parse out and test specially marked time math

### LINKED ISSUE #54: Representation of independent `time` variable in SBML.Model
Currently test cases where the independent variable representing time occurs in the kineticLaw fail (see for instance [case 00852](https://github.com/sbmlteam/sbml-test-suite/blob/52d94baf97a005b6e1fdbdb6116f5c7b4a8a100c/cases/semantic/00852/00852-sbml-l3v2.xml#L69) and [AST_NAME_TIME](http://model.caltech.edu/software/libsbml/5.18.0/docs/formatted/c-api/class_a_s_t_node__t.html)). Is there a chance to parse time to Catalyst.DEFAULT_IV (see suggestion [here](https://github.com/paulflang/SBML.jl/blob/ea554218a17731cbdf1c2f189c04ac3bcc39e652/src/math.jl#L26). Or even better, create a SBML.jl native default_iv so that the Catalyst dependency can be removed from SBML.jl when we create an own repo for the MTK converter)?

exaexa:
Closes #54

We can eventually ask someone about what's the expected overlap of the Time vs Name features in SBML. This version should kinda handle all of it; if it gets simpler, we can erase the `id` field from time.

### DIFF
--- a/src/math.jl
+++ b/src/math.jl
@@ -24,7 +24,16 @@ pointer to `ASTNode_t`.
 """
 function parse_math(ast::VPtr)::Math
     if ast_is(ast, :ASTNode_isName)
-        return MathIdent(get_string(ast, :ASTNode_getName))
+        if ccall(sbml(:ASTNode_getType), Cint, (VPtr,), ast) == 262
+            # This is a special case checking for the value of "simulation
+            # time" as defined by SBML. The constant `262` is the value of the
+            # enum AST_NAME_TIME in `libsbml/src/sbml/math/ASTNodeType.h`,
+            # needs to be kept up to date with the library (otherwise this
+            # breaks).
+            return MathTime(get_string(ast, :ASTNode_getName))
+        else
+            return MathIdent(get_string(ast, :ASTNode_getName))
+        end
     elseif ast_is(ast, :ASTNode_isConstant)
         return MathConst(get_string(ast, :ASTNode_getName))
     elseif ast_is(ast, :ASTNode_isInteger)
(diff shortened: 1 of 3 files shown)


paulflang (commented):
I am wondering if we should merge MathIdent, MathVal, etc. into one single type MathVal with two attributes: `val` and `type`. The latter is filled with what ASTNode_getType returns (or the string version of it, e.g. `AST_INTEGER`, `AST_NAME`, `AST_NAME_TIME`. The former is filled with Strings, Ints, Floats, etc. depending on what `ast_is` returns. `ast_is` would be rewritten as:
```
ast_is(ast::VPtr)::Int = ccall(sbml(:ASTNode_getType), Cint, (VPtr,), ast)
```
This would allow us to represent all the AST node types in SBML.

exaexa:
That "would work" but the explicit type reification a remarkably bad idea for representing ASTs. As an immediate downside, you can't use whatever dispatch to distinguish between data types, and instead need to remember to add sufficient amount of `if`s everywhere. Which is just ugly.

Btw I don't think there's much more node types to cover (check out the header file).

paulflang:
What do you think of using parametric types MathVal{T}?

But I just checked, there is really not much else to cover. Most falls under the MathApply category anyway.

exaexa:
The extra parametrization is be unnecessary here, now we have MathX and MathY, which you would basically convert to MathVal{X} and MathVal{Y}, only adding wrapper layer everywhere.

PR #64, opened 2021-06-17
add test cases for loading dynamic models and convert.jl

### LINKED ISSUE #56: Find and add good tests for loading converters

_Originally posted by @paulflang in https://github.com/LCSB-BioCore/SBML.jl/issues/55#issuecomment-855837778_

paulflang:
I tried to stay close to loadmodel.jl
The `expandInitialAssignments` is not yet working. Maybe Mose can help here.
This is to fix #56

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,7 +14,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 julia = "1.5"
-Symbolics = "1.1"
+Symbolics = "0.1.21"
 
 [extras]
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"


src/converters.jl
```diff
@@ -24,7 +24,7 @@ A converter that runs the SBML conversion routine, with specified conversion
 options. The argument is a vector of pairs to allow specifying the order of
 conversions.
 """
-libsbml_convert(conversion_options::Vector{Pair{String,Dict{String,String}}}) =
+libsbml_convert(conversion_options::Vector{Pair{String,Dict{Any,Any}}}) =  # @MK: [(string(k), string(v), k, v) for (k, v) in Dict()] yields NTuple{4, Any}[]
```
exaexa, line 27: why this? The code relies on the fact that the content of the dictionary is strings.

exaexa: (line 34,35 below is going to fail on any other type)

test/loaddynamicmodels.jl
```diff
@@ -0,0 +1,85 @@
...
+        "https://raw.githubusercontent.com/sbmlteam/sbml-test-suite/master/cases/semantic/00037/00037-sbml-l3v2.xml",
+        "074f0caeaf2cdc390967bfdf06d80ccad519c648df7f421dc4e9c69e71551dff",
+        SBML.libsbml_convert("expandInitialAssignments"),
+        2,
+        ("reaction1", SBML.MathApply("*", [SBML.MathApply("*", [compartment, k2]), S1])),
+        ("S1", 0.01125)
+    ),
+]
+
+@testset "Loading of models from sbml_test_suite" begin
```
exaexa, line 60: this is cool :]

test/runtests.jl
```diff
@@ -5,7 +5,8 @@ using SBML: Model, Reaction, Species
 import Pkg
 
 @testset "SBML test suite" begin
-    include("version.jl")
-    include("ecoli_flux.jl")
-    include("loadmodels.jl")
+    # include("version.jl")
+    # include("ecoli_flux.jl")
+    # include("loadmodels.jl")
+    include("loaddynamicmodels.jl")
```
exaexa, line 11: pls uncomment

exaexa (changes requested):

src/converters.jl
paulflang: @exaexa : I did this because e.g. libsbml_convert("promoteLocalParameters") fails with a method error (complaining that there is not method `libsbml_convert(conversion_options::Vector{Pair{String,Dict{Any,Any}}})` if I remember correctly). This is because `libsbml_convert([converter => Dict(string(k) => string(v) for (k, v) in kwargs)])` produces a `Dict{Any,Any}` `kwargs` is an empty Dict.

exaexa: ah okay, then we need to type it better on the call site I'd say. I'll patch the call.

```julia
Dict{String,String}(string(k) => string(v) ............. )
```

PR #73, opened 2021-06-21
Strip whitespaces when reporting error messages from `readSBML`

giordano:
Also:

* replace `@warn` with `@error`
* actually test the logging message in `readSBML`, instead of showing an extra
  message saying to ignore it.

---

_Old question_:

Related, you may want to use `@error` instead of `@warn` to report the SBML
error?  Or you don't want to scare users too much?

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -99,8 +99,8 @@ function readSBML(fn::String, sbml_conversion = document -> nothing)::SBML.Model
         n_errs = ccall(sbml(:SBMLDocument_getNumErrors), Cuint, (VPtr,), doc)
         for i = 0:n_errs-1
             err = ccall(sbml(:SBMLDocument_getError), VPtr, (VPtr, Cuint), doc, i)
-            msg = get_string(err, :XMLError_getMessage)
-            @warn "SBML reported error: $msg"
+            msg = strip(get_string(err, :XMLError_getMessage))
+            @error "SBML reported error: $msg"
         end
         if n_errs > 0
             throw(AssertionError("Opening SBML document has reported errors"))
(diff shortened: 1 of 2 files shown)


exaexa:
re whitespace stripping, is there some test case that would fail without it? I'm not perfectly sure if whitespace is significant or not (it's used for actual machine identifiers of stuff, so it technically may be... :( ).

re the errors, yeah, that is likely a good idea. My original idea was that the error is going to come after anyway (some of the reported SBML errors are actually just warnings). It should be possible to make it "right" by testing the severity of all reported SBML errors, but I didn't really need to invest the time into that yet. :D

giordano:
> re whitespace stripping, is there some test case that would fail without it? I'm not perfectly sure if whitespace is significant or not (it's used for actual machine identifiers of stuff, so it technically may be... :( ).

I started this PR by adding the log test, and it failed because the logged message was `"SBML reported error: File unreadable.\n"`, which is kinda useless.  To be safe, I can move the `strip` from `get_string` to where `@warn` is used, and leave `get_string` as is (so other uses of this function aren't affected) what do you think?

> re the errors, yeah, that is likely a good idea. My original idea was that the error is going to come after anyway (some of the reported SBML errors are actually just warnings). It should be possible to make it "right" by testing the severity of all reported SBML errors, but I didn't really need to invest the time into that yet. :D

Ok, then I'm going to replace `@warn` with `@error` then :slightly_smiling_face:

giordano:
Ok, I've updated the pull request to only trim whitespaces in the logging message reporting the read error.  `get_string` is left unchanged.

exaexa (approved):

paulflang:
We may want to think about throwing a warning after the conversion in line 109, too. I would not throw an error there, cause most issues arising during conversion are not that severe. But I think libSBML has different degrees of severity with conversion problems. Perhaps Sarah can recommend how to best handle them.

PR #74, opened 2021-06-21
Remove `DocumenterTools` from `docs/` environment

giordano:
The package doesn't seem to be actually used, as far as I could see.


exaexa:
OK, I guess I planned something and it didn't materialize. Thanks!

PR #75, opened 2021-06-21
Run CI with GitHub Actions on pull requests

giordano:
I noticed CI doesn't run on pull requests, I assume this was simply overlooked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-    pull_request:
+  pull_request:
 
 jobs:
   test:


exaexa:
cc @laurentheirendt, any opinion on this?
(we originally didn't want to overload github b/c the tests were ran locally at our gitlab etc, but I guess it makes sense now)

giordano:
Any particular reason why CI runs _only_ on Julia nightly?  That's an... interesting... choice :sweat_smile:  I usually test on the oldest and most recent supported versions of Julia.  Here you declare compatibility with Julia v1.0, but I don't think that's true, since JLL packages require at least Julia v1.3

laurentheirendt:
good catch 👍🏼  will need to include runs on Julia 1.5 and Julia 1.6, not only `latest` 😄

PR #76, opened 2021-06-21
Gitignore files downloaded during the tests

giordano:


giordano:
Also, how about downloading these files into `test/data`?  I'm not a big fan of having all these files around in the directory where is the source code.  I'm annoying, I know :grin:

laurentheirendt:
you are right - feel free to change 👍🏼

PR #78, opened 2021-06-21
Make `convert_simplify_math` constant

giordano:


exaexa:
Ok I should probably apply this to all tiny functions that I'm defining this way. Didn't realize the `const` can help it.

PR #79, opened 2021-06-21
Make `default_symbolics_mapping` constant

giordano:


exaexa:
Good point. I wondered the users might eventually modify this dynamically for convenience, but that'd actually be more dangerous then useful I guess.

giordano:
In general `const` doesn't mean constant value, but constant type.  You can mutate a mutable `const` container, like a `Dict`:
```julia
julia> SBML.default_symbolics_mapping["cos"] = :cos
:cos

julia> SBML.default_symbolics_mapping
Dict{String, Any} with 13 entries:
  "piecewise" => :(Core.ifelse)
  "leq"       => :<=
  "lt"        => :<
  "power"     => :^
  "gt"        => :>
  "floor"     => :floor
  "-"         => :-
  "/"         => :/
  "+"         => :+
  "*"         => :*
  "geq"       => :>=
  "ceiling"   => :ceil
  "cos"       => :cos
```

PR #84, opened 2021-06-22
fun accessors for initial amounts and concentrations

exaexa:
cc @paulflang

### DIFF
--- a/test/ecoli_flux.jl
+++ b/test/ecoli_flux.jl
@@ -18,7 +18,9 @@ end
 
     @test typeof(mdl) == Model
 
-    @test_logs (:error, "SBML reported error: File unreadable.") @test_throws AssertionError readSBML(sbmlfile * ".does.not.really.exist")
+    @test_logs (:error, "SBML reported error: File unreadable.") @test_throws AssertionError readSBML(
+        sbmlfile * ".does.not.really.exist",
+    )
 
     @test length(mdl.compartments) == 2
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -52,3 +52,75 @@ Extract the vector of objective coefficients of each reaction.
 function getOCs(m::SBML.Model)::Vector{Float64}
     return broadcast(x -> x.oc, values(m.reactions))
 end
+
+"""
+    initial_amounts(m::SBML.Model; convert_concentrations = false)
+
+Return a vector of initial amounts for each species, or `nothing` if the
+information is not available. If `convert_concentrations` is true and there is
+information about initial concentration available together with compartment
+size, the result is computed from the species' initial concentration.
+
+In the current version, units of the measurements are completely ignored.
+"""
+function initial_amounts(
+    m::SBML.Model;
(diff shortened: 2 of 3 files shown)


paulflang (commented):
I think this could work with a minor modification to SBMLToolkit and 2 minor modifications here:

- Shouldn't it be `m.species` instead of `m.metabolites`?
- Since the m.metabolites is a `Dict` (i.e. unordered), I think the return vector will be in some weird order. Can we return a vector of `metabolite.id` => initial_amount `Pairs` (that type would be identical to what MTK needs, except that I have to convert the `metabolite.id` from `String` to a time dependent Symbolics FnType).

paulflang:
OK. A second thought on the types I had when thinking about how to do the analogous four kinetic_math. Do you think `initial_amount` should return the same type as `model.metabolites`, i.e. a `Dict(String, Species)`? Or shall we do `Dict(String, Real)` instead of the Vector of Pairs I suggested above?

Similarly do you think the `to_extensive_math()` function (or whatever more appropriate name we come up with) should return the same type as `model.reactions`, i.e. a `Dict(String, Reaction)`? Or shall we do `Dict(String, Math)` or `Vector{Pair{String,Math}}`?

anandijain:
```julia
julia> subtypes(AbstractDict)
9-element Vector{Any}:
 Base.EnvDict
 Base.ImmutableDict
 Base.Iterators.Pairs
 Dict
 IdDict
 OrderedCollections.LittleDict
 OrderedCollections.OrderedDict
 Test.GenericDict
 WeakKeyDict
```
It looks like a Pair iterator subtypes AbstractDict, so you may be able to keep it general

exaexa:
OK I guess I can return a Dict, should be better. I'll look into returning actual generators later, with all other types.

test/ecoli_flux.jl
```diff
@@ -18,7 +18,9 @@ end
 
     @test typeof(mdl) == Model
 
-    @test_logs (:error, "SBML reported error: File unreadable.") @test_throws AssertionError readSBML(sbmlfile * ".does.not.really.exist")
+    @test_logs (:error, "SBML reported error: File unreadable.") @test_throws AssertionError readSBML(
+        sbmlfile * ".does.not.really.exist",
```
anandijain, line 22: nitpick but why is the error message `.does.not.really.exist` with the periods?

exaexa: that's not the mesage.

anandijain: ok u can resolve this then

anandijain: oh yeah totally misread that

paulflang:
In the interest of time it might make sense to return a Dict rather than a generator?

anandijain:
Yeah can we get this in soon, i really don't mind Dict

exaexa:
@paulflang @anandijain the functions now return generators, you can easily get any type you want. See docs of `initial_amounts` for an example with dictionaries.

PR #88, opened 2021-06-25
fix getInitialConcentrations

### LINKED ISSUE #87: Fix getInitialConcentration
https://github.com/LCSB-BioCore/SBML.jl/blob/1e9cdb7aa18cebcefdb75773dcb1ac05c7e23279/src/readsbml.jl#L250

In this line we need to `getInitialConcentration` rather than `getInitialAmount`.

paulflang:
fix #87


anandijain:
lol beat me to it

anandijain:
but ideally we would probably want a test case that would have shown this error before it got merged.

exaexa (approved):
OK I have no idea how this bug actually appeared again, I'm pretty sure I fixed it at some point, somewhere. Unbelievable. :sweat_smile:

Issue #92, opened 2021-06-26
Stable docs don't seem to be updated


exaexa:
This is most likely for @laurentheirendt -- I just noticed that the docs here https://lcsb-biocore.github.io/SBML.jl/stable/ point to a really historic version; same thing at `dev/` looks updated though.Stable link seems pointing to 0.4.0 now...

I kinda hope I didn't break anything back then. :D

laurentheirendt:
I think this can be closed now ...

PR #93, opened 2021-06-26
Add SSH key to let TagBot trigger building documentation

giordano:
Ref #92.  However after merging this you'll need to manually trigger the rebuild
of the missing version, see the "Fixing broken release deployments" box in
https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#Documentation-Versions


exaexa:
Wow, cool, thanks!

@laurentheirendt if you'd have a second, can you check if this will work with our gitlab integration?

Anyway, we've had fun issues with documentation elsewhere, perhaps this is the missing piece :D

laurentheirendt (approved):

PR #94, opened 2021-06-27
add ln, exp, to symbolics.jl

anandijain:
@exaexa is there a place we can view all the names and add more? we are probably missing a bunch


exaexa:
> @exaexa is there a place we can view all the names and add more? we are probably missing a bunch

SBML documentation!

Anyway the list can be fed dynamically to the conversion function (even rewritten inplace) so I guess we don't need to release RN, right? :D

exaexa (approved):

PR #95, opened 2021-06-27
bump Symbolics

paulflang:


Project.toml
```diff
@@ -14,7 +14,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 julia = "1.5"
-Symbolics = "0.1.21"
+Symbolics = "1"
```
exaexa, line 17: does this mean SBML.jl actually cannot run with version 0.1.21?

exaexa (changes requested):

Project.toml
exaexa: I guess we should support both "0.1.21, 1"

exaexa: (also I'm kinda expecting that Julia tooling is going to complain about wildcarding all 1.* releases when the release roll is not finished there yet :D)

paulflang: yes. sorry. will do.

giordano: > (also I'm kinda expecting that Julia tooling is going to complain about wildcarding all 1.* releases when the release roll is not finished there yet :D)

Why?

exaexa: After re-reading the doc section on compat, I guess this should work best:
```julia
Symbolics = "0, ~1.0, ~1.1, ~1.2"
```

giordano: Why?  "1" is perfectly fine

exaexa: future "minor" releases are supposed to be backwards compatible? If so, "1" is all right

paulflang: I think we want the left 0.1.21 <= x < 2, which is "0.1.21, 1", right?

giordano: > future "minor" releases are supposed to be backwards compatible? If so, "1" is all right

By definition, minor releases in the x.y series, with x >= 1, don't introduce breaking changes, provided that the package developers follow semantic versioning.  If you stick to the public API of a package, all versions within a specific major version should be backward compatible.

> I think we want 0.1.21 <= x < 2, which is "0.1.21, 1", right?

It isn't right, because in the 0.x series each minor version is breaking: https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0

paulflang: But apparently it does not break SBML.jl, is it? So why shouldn't SBML.jl allow either?

exaexa: ok so it seems the current version is in fact okay -- supported since 0.1.21, no breaking changes allowed. Thanks!

giordano: Because the portion of public API used (well, tested at least) here didn't change.  But I suppose some other breaking changes have been introduced between 0.1 and 1.0, it just happens SBML wasn't touched by them.

exaexa (approved):

PR #106, opened 2021-07-02
Improve reporting of errors in documents

giordano:
This rationalise the way we report errors from libsbml in a few different
places.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SBML"
 uuid = "e5567a89-2604-4b09-9718-f5f78e97c3bb"
 authors = ["Mirek Kratochvil <miroslav.kratochvil@uni.lu>", "LCSB R3 team <lcsb-r3@uni.lu>"]
-version = "0.5.4"
+version = "0.5.5"
 
 [deps]
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
--- a/src/SBML.jl
+++ b/src/SBML.jl
@@ -7,10 +7,10 @@ using Symbolics
 include("structs.jl")
 include("version.jl")
 
+include("utils.jl")
 include("readsbml.jl")
 include("converters.jl")
 include("math.jl")
-include("utils.jl")
 include("symbolics.jl")
 
 sbml(sym::Symbol) = dlsym(SBML_jll.libsbml_handle, sym)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,3 +1,5 @@
+const VPtr = Ptr{Cvoid}
+
 """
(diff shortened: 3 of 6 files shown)


src/SBML.jl
```diff
@@ -7,10 +7,10 @@ using Symbolics
 include("structs.jl")
 include("version.jl")
 
+include("utils.jl")
```
exaexa, line 10: I was kinda thinking to move the VPtr and other types to a separate types.jl already; there might be more coming. Let's keep that in mind.

src/utils.jl
```diff
@@ -130,3 +132,34 @@ function extensive_kinetic_math(m::SBML.Model, formula::SBML.Math)
...
+"""
+    check_errors(success::Integer, doc::Ptr{Cvoid}, error::Exception)
+
+If success is a 0-valued `Integer` (a logical `false`), then call
+[`get_error_messages`](@ref) to show the error messages reported by SBML in the
+`doc` document and throw the `error` if they are more than 1.  `success` is
+typically the value returned by an SBML C function operating on `doc` which
+returns a boolean flag to signal a successful operation.
+"""
+check_errors(success::Integer, doc::VPtr, error::Exception) =
```
exaexa, line 164: Sad that SBML doesn't hold the C convention with 0 = no errors, 12345... = some error with code right away. :]

src/SBML.jl
exaexa: (other types = now basically Maybe)

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "SBML"
 uuid = "e5567a89-2604-4b09-9718-f5f78e97c3bb"
 authors = ["Mirek Kratochvil <miroslav.kratochvil@uni.lu>", "LCSB R3 team <lcsb-r3@uni.lu>"]
-version = "0.5.4"
+version = "0.5.5"
```
exaexa, line 4: please keep the version bumps separate from other commits. Patch releases are okay but chances are there might be more discussion related just to the version numbering (people depend on this).

exaexa (approved):

Project.toml
giordano: The goal of including the version bump in the pull request is to allow registering the new release immediately after: [release early, release often](https://en.wikipedia.org/wiki/Release_early,_release_often) 🙂 it's also much easier to judge within a pull request whether changes are breaking or not, or whether they introduce new publicly facing features, instead of looking back to a bunch of unrelated changesets.

src/utils.jl
giordano: True, but I guess the problem with that would be that the list of error codes becomes pretty awful (e.g. without any logical order) pretty quickly, if you plan to add new errors in the future. Also, downstream users would need to keep an up-to-date mapping of codes->errors. The used approach lets you add as many errors as you like (they seem to have quite a few)

PR #109, opened 2021-07-09
Use `Unitful` for units, instead of a custom struct

giordano:
Note: this probably needs some more testing: wherever there is `Dict{String,Quantity}` there should probably be `Dict{String,Number}` because units can be dimensionless and we only get `Float64` or `Int`

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -200,3 +200,57 @@ check_errors(
     error::Exception,
     report_severities = ["Fatal", "Error"],
 ) = Bool(success) || get_error_messages(doc, error, report_severities)
+
+# NOTE: this mapping is valid for Level 3/Version 2, it *may* not be valid for
+# other versions
+const KIND_TO_UNIT = Dict(
+    0  => 1.0 * u"A", # UNIT_KIND_AMPERE
+    1  => ustrip(u"mol^-1", Unitful.Na), # UNIT_KIND_AVOGADRO
+    2  => 1.0 * u"Bq", # UNIT_KIND_BECQUEREL
+    3  => 1.0 * u"cd", # UNIT_KIND_CANDELA
+    4  => 1.0 * u"°C", # UNIT_KIND_CELSIUS
+    5  => 1.0 * u"C", # UNIT_KIND_COULOMB
+    6  => 1, # UNIT_KIND_DIMENSIONLESS
+    7  => 1.0 * u"F", # UNIT_KIND_FARAD
+    8  => 1.0 * u"g", # UNIT_KIND_GRAM
+    9  => 1.0 * u"Gy", # UNIT_KIND_GRAY
+    10 => 1.0 * u"H", # UNIT_KIND_HENRY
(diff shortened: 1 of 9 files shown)


src/utils.jl
```diff
@@ -161,3 +161,57 @@ returns a boolean flag to signal a successful operation.
...
+    13 => 1.0 * u"J", # UNIT_KIND_JOULE
+    14 => 1.0 * u"kat", # UNIT_KIND_KATAL
+    15 => 1.0 * u"K", # UNIT_KIND_KELVIN
+    16 => 1.0 * u"kg", # UNIT_KIND_KILOGRAM
+    17 => 1.0 * u"L", # UNIT_KIND_LITER
+    18 => 1.0 * u"L", # UNIT_KIND_LITRE
+    19 => 1.0 * u"lm", # UNIT_KIND_LUMEN
+    20 => 1.0 * u"lx", # UNIT_KIND_LUX
+    21 => 1.0 * u"m", # UNIT_KIND_METER
+    22 => 1.0 * u"m", # UNIT_KIND_METRE
```
exaexa, line 190: :sob:

src/utils.jl
```diff
@@ -161,3 +161,57 @@ returns a boolean flag to signal a successful operation.
 """
 check_errors(success::Integer, doc::VPtr, error::Exception) =
     Bool(success) || get_error_messages(doc, error)
+
+# NOTE: this mapping is valid for Level 3/Version 2, it *may* not be valid for
+# other versions
```
exaexa, line 166: I was wondering if we wouldn't be "safer" with using SBML to convert the unit ID to string (unit string names are imho more stable than enums) and only map from Strings to Units here. I had that with the original units (ccall `UnitKind_toString`).

exaexa (approved):
Looks great! I've left 1 comment for about string vs enum unit kinds. I'm OK with merging the enum variant, just pls ack that the stringish variant was considered. :]

exaexa:
Scheduled for 0.6.0 (releasing 0.5.6 now for easier SBMLTK dev)

src/utils.jl
giordano: Done!

test/data/units.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
giordano, line 1: I wanted to test a couple of specific features of reading units, so I crafted a simple model myself.

When we'll have the possibility to write models to disk we can delete this file and generate it dynamically (or maybe do everything in buffers, without passing through the disk)

giordano (commented):
Ok, this should now be ready for review, I fixed dimensionless units

test/data/units.xml
exaexa: I guess this is perfectly okay now

exaexa (approved):

exaexa:
perfect, thanks a lot!

cc @paulflang

PR #114, opened 2021-07-10
fix argument order and count for piecewise function

### LINKED ISSUE #112: piecewise handling fails
[case 000190](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/00190/00190-sbml-l3v2.xml) gives me the following error:

```
ERROR: LoadError: TypeError: non-boolean (Sym{Real, Nothing}) used in boolean context
Stacktrace:
  [1] (::SBML.var"#conv#31"{Dict{String, Any}, SBML.var"#27#32", SBML.var"#28#33"})(x::SBML.MathApply)
    @ SBML C:\Users\wolf5212\.julia\packages\SBML\ISPz3\src\symbolics.jl:106
  [2] convert(::Type{Num}, x::SBML.MathApply; mapping::Dict{String, Any}, convert_time::SBML.var"#27#32", convert_const::SBML.var"#28#33")
    @ SBML C:\Users\wolf5212\.julia\packages\SBML\ISPz3\src\symbolics.jl:112
  [3] convert(::Type{Num}, x::SBML.MathApply)
    @ SBML C:\Users\wolf5212\.julia\packages\SBML\ISPz3\src\symbolics.jl:106
  [4] mtk_reactions(model::SBML.Model)
    @ SBMLToolkit c:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\SBMLToolkit\src\reactionsystem.jl:97
```
In #40 we used
```julia
function parse_piecewise(val, cond, other)
    IfElse.ifelse(cond, val, other)
end
```
But could also have to do with handling of Booleans.
cc @anandijain

exaexa:
Fixes #112

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,14 +4,15 @@ authors = ["Mirek Kratochvil <miroslav.kratochvil@uni.lu>", "LCSB R3 team <lcsb-
 version = "0.5.5"
 
 [deps]
+IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 SBML_jll = "bb12108a-f4ef-5f88-8ef3-0b33ff7017f1"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 
 [compat]
-julia = "1.5"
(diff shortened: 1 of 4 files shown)


paulflang (approved):
Thanks Mirek. I am not 100% sure how libSBML handles "piecewise" with > 2 pieces, but if you always get the 1+2k arguments that you mentioned in the issue I think this should work.

PR #117, opened 2021-07-13
handle SBML error severities (and don't throw on warnings)

exaexa:


exaexa:
oh dear I accidentally closed this
the computers, all buttons are so close

exaexa:
@anandijain can you check that your models don't explode with the parameter's units warnings with this and ack the review if it's okay?

anandijain:
the issue is that it does one for each which is really excessive imo. but i get that it makes sense to report each of them but i personally do not care. maybe just reporting like one of each (a variable is missing a unit, a parameter is missing a unit...)
but not one for each.

but yes I'm testing it now

exaexa:
> maybe just reporting like one of each

This just copies SBML behavior now, I don't know if there's any functionality to simplify the warning output. May be interesting to dig for it later.

anandijain:
yeah i get an absolute barrage of warnings and errors. filling my whole 10k lines of terminal output almost instantly

exaexa:
any suggested solution? perhaps:
- flag for not reporting the warnings
- throttle at ~100 warnings/infos

giordano:
Instead of flag + throttle, maybe have a threshold of warnings/errors to report?

paulflang (approved):

PR #120, opened 2021-07-15
Rename `SBMLVersion` -> `Version` and unexport it

giordano:
Since we're going to have a breaking version, here is another breaking change.


exaexa (approved):
Looking at the export list we could unexport even the `getS` and other helpers. And maybe rename them to something a bit less.......overshortened. :D

giordano:
> Looking at the export list we could unexport even the getS and other helpers. And maybe rename them to something a bit less.......overshortened. :D

Sounds good to me.  Maybe open an issue so that we can discuss and list the new names?

PR #123, opened 2021-07-16
Add function to read SBML model from a string

### LINKED ISSUE #111: `parseSBML` function to directly pass in an XML string
it would work identically to readSBML, but the string of the model is in RAM, not saved to disk. It is annoying to have to save then read sometimes

giordano:
Fix #111.  CC: @anandijain

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -71,38 +71,13 @@ function get_optional_double(x::VPtr, is_sym, get_sym)::Maybe{Float64}
     end
 end
 
-"""
-    readSBML(
-        fn::String,
-        sbml_conversion = document -> nothing;
-        report_severities = ["Fatal", "Error"],
-    )::SBML.Model
-
-Read the SBML from a XML file in `fn` and return the contained `SBML.Model`.
-
-The `sbml_conversion` is a function that does an in-place modification of the
-single parameter, which is the C pointer to the loaded SBML document (C type
-`SBMLDocument*`). Several functions for doing that are prepared, including
-[`set_level_and_version`](@ref), [`libsbml_convert`](@ref), and
-[`convert_simplify_math`](@ref).
-
-`report_severities` switches on and off reporting of certain errors; see the
-documentation of [`get_error_messages`](@ref) for details.
-
-# Example
-```
-m = readSBML("my_model.xml", doc -> begin
-    set_level_and_version(3, 1)(doc)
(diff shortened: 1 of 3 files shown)


src/readsbml.jl
```diff
@@ -71,6 +71,34 @@ function get_optional_double(x::VPtr, is_sym, get_sym)::Maybe{Float64}
     end
 end
 
+function _readSBML(
+    symbol::Symbol,
+    fn::String,
+    sbml_conversion = document -> nothing,
+    report_severities = ["Fatal", "Error"],
```
exaexa, line 78: this probably doesn't need the defaults but I guess it's okay

(is there a good julia way to "forward" the fact that the argument is missing, apart form `Missing` ?)

src/readsbml.jl
```diff
@@ -97,32 +127,37 @@ m = readSBML("my_model.xml", doc -> begin
...
+        report_severities = ["Fatal", "Error"],
+    )::SBML.Model
 
-        if 0 == ccall(sbml(:SBMLDocument_isSetModel), Cint, (VPtr,), doc)
-            throw(AssertionError("SBML document contains no model"))
-        end
+Read the SBML from the string `str` and return the contained `SBML.Model`.
 
-        model = ccall(sbml(:SBMLDocument_getModel), VPtr, (VPtr,), doc)
+The `sbml_conversion` is a function that does an in-place modification of the
```
exaexa, line 145: avoid copypastes, just refer that "the other arguments work as with `readSBML`"

test/loadmodels.jl
```diff
@@ -58,7 +58,7 @@ sbmlfiles = [
     ),
 ]
 
-@testset "Loading of models from various sources" begin
+@testset "Loading of models from various sources - $(reader)" for reader in (readSBML,  readSBMLFromString)
```
exaexa, line 61: lovely.

src/readsbml.jl
giordano: Done!

src/readsbml.jl
giordano: Oh, yeah, this internal function doesn't really need a default, I'm going to remove it.

> (is there a good julia way to "forward" the fact that the argument is missing, apart form Missing ?)

I guess the julian way would be a different method.

exaexa (approved):

Issue #125, opened 2021-07-17
expandInitialAssignments converter fails on case 01234


paulflang:
## Minimal code example to reproduce the problem
```julia
using SBMLToolkit
using Symbolics

case = "01234"
url = "https://raw.githubusercontent.com/sbmlteam/sbml-test-suite/master/cases/semantic/$case/$case-sbml-l3v2.xml"
tmpfile = download(url)
try
    SBMLToolkit.checksupport(tmpfile)
    ml1 = readSBML(tmpfile)  # works
    ml2 = readSBML(tmpfile, doc -> begin
            SBMLToolkit.SBML.libsbml_convert("expandInitialAssignments")(doc)
        end)  # fails
finally
    rm(tmpfile)
end
```

## Expected behavior

`ml1` and `ml2` should be the same, because the initialAssignment contains no mathML and according to the [test case descriptions](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01234/01234-model.m) should therefore be ignored. It could be that I am doing sth wrong or the problem is rooted in SBML.jl or in the libSBML C API. At least setting `report_severities = []` does not help (if it has to do with the libSBML API we should report this there @giordano )

## Actual behavior
```
ERROR: LoadError: InexactError: Bool(-3)
Stacktrace:
 [1] Bool
   @ .\float.jl:106 [inlined]
 [2] check_errors (repeats 2 times)
   @ C:\Users\wolf5212\.julia\packages\SBML\XMU1n\src\utils.jl:202 [inlined]
 [3] (::SBML.var"#3#4"{Vector{Pair{String, Dict{String, String}}}})(doc::Ptr{Nothing})
   @ SBML C:\Users\wolf5212\.julia\packages\SBML\XMU1n\src\converters.jl:48
 [4] (::var"#37#38")(doc::Ptr{Nothing})
   @ Main c:\Users\wolf5212\OneDrive - Nexus365\DPhil Project\Programmes\juliadev\dev\Sandbox\mwe_case00851.jl:12
 [5] readSBML(fn::String, sbml_conversion::var"#37#38"; report_severities::Vector{String})
   @ SBML C:\Users\wolf5212\.julia\packages\SBML\XMU1n\src\readsbml.jl:113
 [6] readSBML(fn::String, sbml_conversion::Function)
   @ SBML C:\Users\wolf5212\.julia\packages\SBML\XMU1n\src\readsbml.jl:105
 [7] top-level scope
[...]

PR #127, opened 2021-07-17
Always pass `report_severities` argument to `check_errors`

giordano:
Related to #125, we weren't forwarding `report_severities` to `check_errors`, so
there was no way to actually see the warning.

Note: this depends on #126, but I opened it as a separate pull request because I
think it's logically independent from that one.

### DIFF
--- a/src/converters.jl
+++ b/src/converters.jl
@@ -1,11 +1,12 @@
 
 """
-    set_level_and_version(level, version)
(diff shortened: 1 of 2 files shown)


giordano:
This should be ready for review now

exaexa (approved):
lgtm

PR #137, opened 2021-07-28
docstringextensions

### LINKED ISSUE #136: DocStringExtensions usage (particularly on structs)
I might do this soon, but want to keep an issue open so I don't forget.

anandijain:
image

now the struct docstrings I changed look like this

the readme is displayed if you do `?SBML`

closes #136

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
+DocStringExtensions = "0.8"
 IfElse = "0.1"
 Symbolics = "0.1.21, 1"
 Unitful = "1"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Mirek Kratochvil <miroslav.kratochvil@uni.lu>", "LCSB R3 team <lcsb-
 version = "0.6.0"
 
 [deps]
+DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
```
exaexa, line 7: does this need compat?

src/SBML.jl
```diff
@@ -1,5 +1,9 @@
+"""
+$(DocStringExtensions.README)
```
exaexa, line 2: I'm not sure if displaying the whole README is a good idea (it's github-ish), will write some shorter introduction instead.

src/structs.jl
```diff
@@ -77,7 +77,12 @@ struct MathLambda <: Math
 end
 
 """
+$(TYPEDEF)
+
 SBML Compartment with sizing information.
+
+# Fields
+$(FIELDS)
```
exaexa, line 85: is it possible to get the fields together with types?

exaexa:
wow this is cool

PR #141, opened 2021-08-01
rational mathval

### LINKED ISSUE #140: SBML doesn't respect math <cn> rationals
```xml
<kineticLaw>
    <math xmlns="http://www.w3.org/1998/Math/MathML">
        <cn type="rational">1<sep />5000</cn>
    </math>
</kineticLaw>
```
to reproduce just add `<cn type="rational">1<sep />5000</cn>` to any kineticLaw

```julia
m = readSBML(...)
m.reactions["reaction1"].kinetic_math
```
I get a `SBML.MathVal{Float64}(0.0002)` instead of a rational. it looks like the issue is just https://github.com/LCSB-BioCore/SBML.jl/blob/master/src/math.jl#L60 

I tried adding a 
```julia 
    elseif ast_is(ast, :ASTNode_isRational)
        return MathVal(ccall(sbml(:ASTNode_getRational), Cint, (VPtr,), at))
```
While `isRational` exists, sadly: 

>Rational numbers are represented in an AST node using separate numerator and denominator values. These can be retrieved using the methods ASTNode::getNumerator() and ASTNode::getDenominator(). [source](http://sbml.org/Software/libSBML/5.18.0/docs/cpp-api/class_a_s_t_node.html)

So I'm not exactly sure how to fix it, why they don't just have a `getRational` is beyond me.

anandijain:
closes #140 

idk a good way to add a test case for this since `SBML.MathVal(1//2)` already works, but when rationals occur in an SBML file, they get promoted to Float


exaexa:
ey cool, is there a testcase for this? (best just add the SBML testcase you used in #140 to the list in `test/loadmodels.jl`)

anandijain:
I just took case 1 from the suite and added a rational cn in the kineticLaw. I can't test this anymore as my old laptop died now I have m1. 😆 . I might try to build myself soon

anandijain:
https://gist.githubusercontent.com/anandijain/a3f20f56a0e61a943c578e3892d3112a/raw i made a gist that can be `download`ed but I don't know what to put in the `hash` part of the tuple.

and you `@test m.reactions["reaction1"].kinetic_math.val isa Rational`

exaexa:
nah, we cant depend on random stuff uploaded to gist, is there a SBML testcase for this?

anandijain:
all the uses of rationals i found were in `eventassignment`s. i can look in biomd later

PR #142, opened 2021-08-02
separate reactants and product stoichiometries

### LINKED ISSUE #119: Adding `substoich` and `prodstoich` fields to `Reaction`
## Target functionality

Consider for instance reaction1 in [case 00054](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/00054/00054-sbml-l3v2.xml):
S1 + S2 -> 2*S2

When we create the Reaction.stoichiometry Dict, we get `Dict("S1"=>-1," S2"=>1)` .

This makes it impossible to identify further downstream whether the reaction kinetic is mass action. Therefore, I suggest to create two different fields: `substoich` and `prodstoich` in `Reaction`.

## Desired output

I would like to be able to do
```julia
ml = readSBML("case00054.xml")
@test isequal(ml.reactions["reaction1"].substoich, Dict("S1"=>1, "S2"=>1))
@test isequal(ml.reactions["reaction1"].prodstoich, Dict("S2"=>2))
```

## Suggestions for implementation

To make this change non-breaking, the `stoichiometry` field could return the sth like Reaction.prodstoich minus Reaction.substoich.

anandijain:
to close #119

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -358,42 +358,23 @@ function extractModel(mdl::VPtr)::SBML.Model
         end
 
         # extract stoichiometry
-        stoi = Dict{String,Float64}()
-        add_stoi(sr, factor) = begin
+        reactants = Dict{String,Float64}()
+        products = Dict{String,Float64}()
+
+        add_stoi(sr, coll) = begin
             s = get_string(sr, :SpeciesReference_getSpecies)
-            stoi[s] =
-                get(stoi, s, 0) +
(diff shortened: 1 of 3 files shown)


anandijain:
oops ill need to separate this out accidentally grabbed my old commit

exaexa:
I removed the copypasta and unmatlabed the variable names, good to go now.

exaexa:
(wrong button whoops)

PR #143, opened 2021-08-07
Add `writeSBML`

giordano:
This is a very preliminary draft, it only takes units into account

EDIT by @exaexa :
Looking at the "dependency order" of stuff in SBML, I guess the good implementation path is:

- [x] parameters
- [x] units
- [x] Compartment
- [x] GeneProducts (~mostly done, but gene products need to be added to an FBC package, not directly to the model~ now done, but it requires https://github.com/JuliaPackaging/Yggdrasil/pull/5011)
- [x] GeneProductAssociations
- [x] Math (should be mostly complete, needs double checking)
- [x] InitialAssignments
- [x] FunctionDefinitions (depends on Math)
- [x] Rules (depend on Math)
- [x] Species (~almost done, needs writing the formula~ done and fully tested)
- [x] Reactions
- [x] Objectives (requires #211)
- [x] Events
- [x] model metadata (default unit kinds etc)
- [x] whole model


Project.toml
```diff
@@ -9,10 +9,12 @@ IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 SBML_jll = "bb12108a-f4ef-5f88-8ef3-0b33ff7017f1"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+StructHelpers = "4093c41a-2008-41fd-82b8-e3f9d02b504f"
```
giordano, line 12: I guess you aren't going to be happy to _add_ yet another dependency, but the problem is that `==` defaults to `===` for custom structs (see my `#gripes` today in Slack), and so we have to extend all `==` methods for all the structures we want to compare, but this is error prone as we may add new fields to these structs.  `StructHelpers` does this for us automatically.

exaexa: I guess this is good, literally adds just Structhelpers and ConstructionBase (transitively), both of which are tiny

exaexa: (better than writing == instances for everything)

src/writesbml.jl
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    # Add species
+    for (id, species) in mdl.species
+        species_t = ccall(sbml(:Species_create), VPtr, (Cuint, Cuint), 3, 2)
+        ccall(sbml(:Species_setId), Cint, (VPtr, Cstring), species_t, id)
+        isnothing(species.name) || ccall(sbml(:Species_setName), Cint, (VPtr, Cstring), species_t, species.name)
+        isnothing(species.compartment) || ccall(sbml(:Species_setCompartment), Cint, (VPtr, Cstring), species_t, species.compartment)
+        isnothing(species.boundary_condition) || ccall(sbml(:Species_setBoundaryCondition), Cint, (VPtr, Cint), species_t, species.boundary_condition)
+        # isnothing(species.formula) || ccall(sbml(:Species_setFormula), Cint, (VPtr, Cstring), species_t, species.compartment)
```
giordano, line 47: I fear this is related to the problem with gene products: how do we write the formula?  There is no `Species_setFormula` function, and we read from the FBC package: https://github.com/LCSB-BioCore/SBML.jl/blob/8ee591a473a7ecb911c4dff474ca1e8db7337e86/src/readsbml.jl#L295

exaexa: (yeah looks like there's no doxygen around the source)

giordano: Ok, that's useful to know, thanks!  But I think we still don't know how to create the plugin

src/readsbml.jl
```diff
@@ -356,8 +357,11 @@ function get_model(mdl::VPtr)::SBML.Model
             end
             objectives[get_string(o, :Objective_getId)] = Objective(type, flux_objectives)
         end
-        active_objective =
-            get_optional_string(mdl_fbc, :FbcModelPlugin_getActiveObjectiveId)
+        active_objective = let
+            aoi = get_optional_string(mdl_fbc, :FbcModelPlugin_getActiveObjectiveId)
+            # Keep it to nothing if the string is empty
+            isempty(aoi) ? nothing : aoi
+        end
```
giordano, line 364: I wonder if `active_objective` should be `::String` since `FbcModelPlugin_getActiveObjectiveId` returns an empty string, not a null-pointer, when there is no active objective.

exaexa: (solved in slack I guess -- let's make it empty string because it doesn't break anything)

paulflang:
@giordano : would it be hard to merge the new Trigger struct and associated code before this whole PR is done? (Not super urgent, but would be nice to have)

paulflang:
Alright. Then I'll wait and watch this space.

exaexa:
@paulflang please open a PR with the code so that it stays in sight and we can see that we're not doing something that would break the code. Integrity of writeSBML merge is a kinda priority now (partly because how huge this PR is) -- I'll eventually do the gluing&rebasing with the Trigger code later.

exaexa:
Depends on https://github.com/sbmlteam/libsbml/pull/247

exaexa:
Full functionality depends on solving https://github.com/sbmlteam/libsbml/issues/248

exaexa (approved):
:tada: C E L E B R A T I O N :tada: 

I'll eventually need to check whether some of the smaller items that are now created by plain constructor (`Something_create`) wouldn't be safer to get created by parent (inheriting XML props etc, `ParentObject_addSomething` style). But that's a completely minor problem.

Let's roll out 1.0 asap.

PR #144, opened 2021-08-08
Make `sbml` slightly more type-stable

giordano:
`Libdl.dlsym` can return `Union{Nothing, Ptr{Nothing}}`.  We know that the
symbols we use exist, and even if they didn't this change makes the function
fail early instead of calling a `NULL` function in the shared library :boom:


exaexa (approved):
beautiful.

PR #145, opened 2021-08-09
Fix definition of units

giordano:
The exponent applies to the multiplier as well.  Reorganize the `get_unit`
function to make it clearer what's the formula used.

### DIFF
--- a/test/ecoli_flux.jl
+++ b/test/ecoli_flux.jl
@@ -22,7 +22,7 @@ end
         sbmlfile * ".does.not.really.exist",
     )
 
-    @test mdl.units["mmol_per_gDW_per_hr"] == 2.7777e-7 * u"mol * g^-1 * s^-1"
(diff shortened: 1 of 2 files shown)


test/ecoli_flux.jl
```diff
@@ -22,7 +22,7 @@ end
         sbmlfile * ".does.not.really.exist",
     )
 
-    @test mdl.units["mmol_per_gDW_per_hr"] == 2.7777e-7 * u"mol * g^-1 * s^-1"
+    @test mdl.units["mmol_per_gDW_per_hr"] ≈ 3.6001008028224795 * u"mol * g^-1 * s^-1"
```
exaexa, line 25: O_o are the float errors so bad?

giordano: Tests actually pass on my computer with `==`, but don't want to rely on the full 17-digit representation for this to work on all systems. Also, this model is simply wrong 😬

exaexa: I'm scared but merging. :D

PR #146, opened 2021-08-09
add downstream SBMLToolkit testing

anandijain:
now we will see what breaks downstream when changes are made to SBML

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,4 +1,4 @@
-name: IntegrationTest
+name: DownstreamTest
 on:
   push:
     branches: [master]
@@ -31,7 +31,7 @@ jobs:
         with:
           repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
           path: downstream
-      - name: Load this and run the downstream tests
+      - name: Load the tested version of SBML package and run the downstream tests
         shell: julia --color=yes --project=downstream {0}
         run: |
           using Pkg
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -18,6 +18,7 @@ jobs:
         os: [ubuntu-latest]
         package:
           - {user: SciML, repo: SBMLToolkit.jl, group: All}
+          - {user: LCSB-BioCore, repo: COBREXA.jl, group: All}
           
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,9 +1,7 @@
 name: DownstreamTest
 on:
-  push:
-    branches: [master]
-    tags: [v*]
   pull_request:
+    branches: [master]
 
 jobs:
   test:
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,0 +1,50 @@
+name: IntegrationTest
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
+    runs-on: ${{ matrix.os }}
+    env:
+      GROUP: ${{ matrix.package.group }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1]
+        os: [ubuntu-latest]
+        package:
+          - {user: SciML, repo: SBMLToolkit.jl, group: All}
+          
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.julia-version }}
+          arch: x64
+      - uses: julia-actions/julia-buildpkg@latest
+      - name: Clone Downstream
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
+          path: downstream
+      - name: Load this and run the downstream tests
+        shell: julia --color=yes --project=downstream {0}
+        run: |
+          using Pkg
+          try
+            # force it to use this PR's version of the package
+            Pkg.develop(PackageSpec(path="."))  # resolver may fail with main deps
+            Pkg.update()
+            Pkg.test()  # resolver may fail with test time deps
+          catch err
+            err isa Pkg.Resolve.ResolverError || rethrow()
+            # If we can't resolve that means this is incompatible by SemVer and this is fine
+            # It means we marked this as a breaking change, so we don't need to worry about
+            # Mistakenly introducing a breaking change, as we have intentionally made one
+            @info "Not compatible with this release. No problem." exception=err
+            exit(0)  # Exit immediately, as a success
+          end
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -9,6 +9,7 @@ jobs:
   test:
     name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
     env:
       GROUP: ${{ matrix.package.group }}
     strategy:


exaexa:
This looks nice but it would be much better in another repo.

For SBML.jl, this prevents the development from having clean tests on breaking changes. We'd need to switch to develop+master model of merging (master is kept compatible, develop breaks).

Also, it makes the breaking changes unpublishable (you don't pass tests before the upstream fixes them, but upstream can't fix it because the version isn't out yet).

anandijain:
yeah thats a good point, ive seen sciml merge PRs that cause integration tests to fail, but that pass on the repos tests.
although i understand not wanting the red check

mainly its just to know when a change happens that causes something that i need to fix. otherwise i need to be a little more proactive

anandijain:
im overall fine not having it merged, but i don't think it could live in another repo, since it needs to be triggered for changes here

exaexa:
yeah well, now that you have the compat entry there you may just wait for the compat bot that will open a PR which will ping you in case the build is broken.

Regarding the breaking changes, I'll keep letting you know, don't worry. :D

ChrisRackauckas:
It needs to be in the Base repo, that's kind of the point of it being downstream 😅

exaexa:
> It needs to be in the Base repo, that's kind of the point of it being downstream sweat_smile

You guys love cyclic dependencies right? :D

ChrisRackauckas:
There's no dependency. It's just a downstream test.

ChrisRackauckas:
>Regarding the breaking changes, I'll keep letting you know, don't worry. :D

How will you know that a change is breaking if you're not testing the user code to see if it breaks? It ends up just decreasing dev times for PRs to warn for accidental breakage or when the downstream is using something they shouldn't (in which case, the PR helps point that out). The only time you should get a real red X for a downstream test is if you made a break and plan to do a breaking release, in which case it's then a good indicator of ecosystem health (correctly saying that the ecosystem is not keeping up to date).

I mean, the purpose of tests is to help the development process, not to be green. If the purpose of tests was to be green, I could put in a PR to delete all of the tests and it would do a pretty good job 😉

exaexa:
Extra information here--

- merging this would be able to break master in a way we can not fix (except for reverthing this PR), which is not a happy way to do software. Direct pulling of tests to upstream (although lightly here) is also a major source of pain, so we're going to avoid it.
- Instead of this, we'll make a separate integrations repository (or maybe a branch or something, not sure yet) that will also source some other packages that depend on SBML (currently at least cobrexa + sbmltk), build&test on any SBML repo update, and produce the notifications for us&everyone interested.
- The changes in SBML are and will be versioned properly; in turn SBMLTK should not get broken by accident.

Please don't close this yet, I'll pick Anand's commits so that he stays credited in the other repo. ETA is unclear now b/c of other projects, expect mid-September.

(cc @laurentheirendt )

ChrisRackauckas:
>The changes in SBML are and will be versioned properly; in turn SBMLTK should not get broken by accident.

How do you know without testing?

exaexa:
> > The changes in SBML are and will be versioned properly; in turn SBMLTK should not get broken by accident.
> 
> How do you know without testing?

exaexa:
uh wrong button sorry. :D

I'm testing SBML, the tests cover the properties, if the tests need to be fixed, version goes up, that's it.

ChrisRackauckas:
How do you know that your tests are complete enough to find anything that would break downstream usage? It's a pretty heavy assumption that your tests are perfect. Downstream testing makes that no longer an assumption. I think setting it as "allow failures" is fine, but doing no downstream testing means you are really just guessing as to whether there is any downstream breakage.

exaexa:
If there's a property you want to rely on, we should add explicit in-package tests for that. (does julia have any actual property tests?)

Could we rather set up a pipeline trigger to SBMLToolkit from that we'd get notifications? That way at least the tests will run in their own repo. (We're already doing this internally with some packages.)

ChrisRackauckas:
>If there's a property you want to rely on, we should add explicit in-package tests for that. (does julia have any actual property tests?)

That's still developing in a style where you have to believe that you never have and never will make a mistake. That seems like a very strong assumption.

anandijain:
I'm once again asking for this.

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,50 @@
...
+    runs-on: ${{ matrix.os }}
+    env:
+      GROUP: ${{ matrix.package.group }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1]
+        os: [ubuntu-latest]
+        package:
+          - {user: SciML, repo: SBMLToolkit.jl, group: All}
```
exaexa, line 20: is there a straightforward way to extend this for multiple packages? Ideally, can we have LCSB-BioCore/COBREXA.jl  too?

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,50 @@
+name: IntegrationTest
```
exaexa, line 1: name: `Downstream tests`

(this is not integration testing)

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          version: ${{ matrix.julia-version }}
+          arch: x64
+      - uses: julia-actions/julia-buildpkg@latest
+      - name: Clone Downstream
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
+          path: downstream
+      - name: Load this and run the downstream tests
```
exaexa, line 34: load `this` = load what?

(I guess it means "install the tested version of SBML package" ? )

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,50 @@
...
+        shell: julia --color=yes --project=downstream {0}
+        run: |
+          using Pkg
+          try
+            # force it to use this PR's version of the package
+            Pkg.develop(PackageSpec(path="."))  # resolver may fail with main deps
+            Pkg.update()
+            Pkg.test()  # resolver may fail with test time deps
+          catch err
```
exaexa, line 43: Pkg.test() should be outside of the try-catch right? otherwise it won't fail visibly

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
+          path: downstream
+      - name: Load this and run the downstream tests
+        shell: julia --color=yes --project=downstream {0}
+        run: |
+          using Pkg
+          try
+            # force it to use this PR's version of the package
+            Pkg.develop(PackageSpec(path="."))  # resolver may fail with main deps
```
exaexa, line 40: Just checking, the `Pkg.develop()` here loads SBML, and `Pkg.test()` tests the `downstream` because it is a project, right?

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,50 @@
...
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
+    runs-on: ${{ matrix.os }}
+    env:
```
exaexa, line 12: This should make soft fail only. I'm not sure what's the config that would allow it.

We must _never_ get CI failures in master branch/tags that would depend on contents of other repositories.

exaexa: cc @laurentheirendt any ideas about how to conf it? :]

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,50 @@
...
+            Pkg.update()
+            Pkg.test()  # resolver may fail with test time deps
+          catch err
+            err isa Pkg.Resolve.ResolverError || rethrow()
+            # If we can't resolve that means this is incompatible by SemVer and this is fine
+            # It means we marked this as a breaking change, so we don't need to worry about
+            # Mistakenly introducing a breaking change, as we have intentionally made one
+            @info "Not compatible with this release. No problem." exception=err
+            exit(0)  # Exit immediately, as a success
+          end
```
exaexa, line 50: newlineeeeeeeeeeeeeee :D

.github/workflows/downstream.yml
exaexa: (maybe it would be cleaner to avoid the --project and write the paths right away)

(also probably no reason to do update(), the stuff is already installed fresh, right?)

exaexa (changes requested):
some fixes needed

(also we really don't want to merge a failing PR into master)

exaexa:
Also, how come the commit is marked with CI failure but the whole PR is marked as success?

:thinking:

.github/workflows/downstream.yml
anandijain: yes. see https://github.com/SciML/ModelingToolkit.jl/blob/master/.github/workflows/Downstream.yml#L20-L27

anandijain: ill add you guys now

.github/workflows/downstream.yml
anandijain: @ChrisRackauckas im not sure about this one. are you the person to ask?

.github/workflows/downstream.yml
anandijain: what constitutes a "CI failure". I'm fine w merges if the PR passes here but not on the downstream check. CI will still show that as red x. 

@ChrisRackauckas are there settings to avoid it causing the red check? I kind of doubt it

.github/workflows/downstream.yml
anandijain: this is how it is elsewhere . i dont think its a problem

.github/workflows/downstream.yml
exaexa: We might run this in PRs only to provide warning (where it should get green); and not in master to avoid breakage. Is that okay? All code is required to go through PRs anyway.

.github/workflows/downstream.yml
exaexa: Oh wait there's a rethrow below. False alarm, sorry. :D

anandijain:
what still needs to be done?

.github/workflows/downstream.yml
ChrisRackauckas: you can always set it as an allowed failure.

anandijain: okay I think that should do it

exaexa:
> what still needs to be done?

Sorry for delay. Please remove the triggers on the _push_ to master, leaving only PRs to master. I sent some explanation on slacks.

exaexa (approved):
LGTM now

PR #147, opened 2021-08-18
simplify stoichiometry_matrix API & make it more efficient

exaexa:
Writing individual fields to sparse matrices is pretty slow, this should be faster in most cases. More or less a consequence of https://github.com/LCSB-BioCore/COBREXA.jl/pull/443 .

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -272,7 +272,7 @@ function get_unit(u::VPtr)
     exponent = ccall(sbml(:Unit_getExponent), Cint, (VPtr,), u)
     # See page 44 of
     # http://sbml.org/Special/specifications/sbml-level-3/version-2/core/release-2/sbml-level-3-version-2-release-2-core.pdf
-    return (multiplier * unit * exp10(scale)) ^ exponent
+    return (multiplier * unit * exp10(scale))^exponent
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -1,29 +1,48 @@
...
+
+    for (rid, r) in m.reactions
+        ridx = col_idx[rid]
+        for (sid, stoi) in r.reactants
+            push!(SI, row_idx(sid))
+            push!(RI, ridx)
+            push!(SV, -stoi)
+        end
+        for (sid, stoi) in r.products
+            push!(SI, row_idx(sid))
```
giordano, line 40: These are probably
```suggestion
            push!(SI, row_idx[sid])
            push!(RI, ridx)
            push!(SV, -stoi)
        end
        for (sid, stoi) in r.products
            push!(SI, row_idx[sid])
```
?

exaexa: Ah yes I messed up the tests (tested a wrong version and assumed I made it right on the first place). Sorry for premature heads-up :D

exaexa: (should be okay now)

giordano (approved):

PR #148, opened 2021-08-19
Update to Symbolics 3

ChrisRackauckas:
This requires using the safer symbolic variable definition via interpolation.

### DIFF
--- a/src/symbolics.jl
+++ b/src/symbolics.jl
@@ -97,7 +97,10 @@ const default_symbolics_constants =
         ::Type{Num},
         x::SBML.Math;
         mapping = default_symbolics_mapping,
-        convert_time = (x::SBML.MathTime) -> Num(Variable(Symbol(x.id))).val,
+        convert_time = function (x::SBML.MathTime)
+            sym = Symbol(x.id)
+            Symbolics.unwrap(first(@variables \$sym))
+        end,
         convert_const = (x::SBML.MathConst) -> Num(default_symbolics_constants[x.id]),
     )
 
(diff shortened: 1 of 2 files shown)


src/symbolics.jl
```diff
@@ -97,7 +97,10 @@ const default_symbolics_constants =
         ::Type{Num},
         x::SBML.Math;
         mapping = default_symbolics_mapping,
-        convert_time = (x::SBML.MathTime) -> Num(Variable(Symbol(x.id))).val,
+        convert_time = function (x::SBML.MathTime)
+            sym = Symbol(x.id)
+            Symbolics.unwrap(first(@variables \$sym))
```
exaexa, line 102:
```
help?> Symbolics.unwrap
  No documentation found.

  Symbolics.unwrap is a Function.
```

what does this actually do?

ChrisRackauckas: it changes from Num => Sym. `.val` technically works, but it's not a safe way to do it (edge cases like symbolic arrays).

ChrisRackauckas: it should probably get a docstring, agreed (@shashi)

exaexa: OK I'm wondering why we used `.val` in the first place.

Anyway this looks good to go, I'll just need to find a bit of time to get through the possibilities, ideally killing the macro from there :D

Expect merged in a few days.

src/symbolics.jl
```diff
@@ -117,13 +120,19 @@ function Base.convert(
     ::Type{Num},
     x::SBML.Math;
     mapping = default_symbolics_mapping,
-    convert_time = (x::SBML.MathTime) -> Num(Variable(Symbol(x.id))).val,
+    convert_time = function (x::SBML.MathTime)
+        sym = Symbol(x.id)
+        Symbolics.unwrap(first(@variables $sym))
```
shashi, line 125: You can use `variable` function here.

ChrisRackauckas:
Bump

exaexa:
Any news about whether the above (comment from @shashi) can be solved without the macro?

ChrisRackauckas:
This method seemed to make sure the metadata was initialized correctly and worked well in downstream testing. I do not think we should keep the entire ecosystem held back for another week because of a potential syntax bikeshed.

ChrisRackauckas:
If it would make you happy, after merging open an issue about looking at other potential forms for that line of code and I can look at it next week, but for now correctness is my number one priority.

exaexa:
The main problem is that "initialized correctly" is unclear & unspecified here, and there seems to be no piece of documentation in Symbolics saying that the `unwrap(first(@variables unquote(sym)))` combo would be the way (and why); or at least from what I could find now. Perhaps it would be better to just document this methodology in Symbolics.

PR #153, opened 2021-10-10
custom show for SBML.Model

anandijain:
For any large model, the amount of stuff that gets shown to the REPL is overwhelming. 

I found though, that while the `eval(Meta.parse(repr(m))) isa SBML.Model` test works for that model, for another model I tried, it failed. 

This is not an issue introduced by this PR, just something I found.

### DIFF
--- a/test/loadmodels.jl
+++ b/test/loadmodels.jl
@@ -213,9 +213,6 @@ end
 
 @testset "custom show" begin 
     m = readSBML(joinpath(@__DIR__, "data", "custom.xml"))
-    io = IOBuffer()
-    show(io, MIME("text/plain"), m)
-    s = String(take!(io))
-    @test s == "SBML.Model with 1 reactions, 0 species, and 0 parameters."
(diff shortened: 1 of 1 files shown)


exaexa:
Hi! We actually had a similar discussion in COBREXA.jl. This looks pretty good to me.

@stelmo could you please check the implementation? cobrexa implementation is mostly your code, is there anything to add here?

exaexa:
One consideration, can the `show` implementation print out the SBML.Model type verbatim so that people don't need to write `typeof`for finding what's the actual type name if they want to play with it or see the help for it?

anandijain:
Expression: s == "SBML.Model with 1 reactions, 0 species, and 0 parameters."
   Evaluated: "Model with 1 reactions, 0 species, and 0 parameters." == "SBML.Model with 1 reactions, 0 species, and 0 parameters."

I guess Model was directly imported Im guessing

Issue #139, opened 2021-07-31
case 01296 fails


paulflang:
## Minimal code example to reproduce the problem

```julia
case = "01296"
url = "https://raw.githubusercontent.com/sbmlteam/sbml-test-suite/master/cases/semantic/$case/$case-sbml-l3v2.xml"
tmpfile = download(url)
try
    SBMLToolkit.checksupport(tmpfile)
    ml = readSBML(tmpfile, doc -> begin
            set_level_and_version(3, 2)(doc)
            convert_simplify_math(doc)
        end)
    println(ml.reactions["J1"].kinetic_math)
finally
    rm(tmpfile)
end
```

## Expected behavior

Insertion of the rate of change of `p1` in kinetic_math (see specs p25).

## Actual behavior

prints `SBML.MathApply("rateOf", SBML.Math[SBML.MathIdent("p1")])`

paulflang:
A similar problem occurs in 01459

exaexa:
What is the expected SBML math you'd get? (also, if this is really a problem, I don't think we can solve it in SBML.jl, rather open a bug in libsbml)

paulflang:
`SBML.MathApply("*", SBML.Math[SBML.MathIdent("p1"), SBML.MathVal(0.01)])`, bacause the ODE that describes p1 is `p1' = 0.01*p1`. I do not see a reason why this could not be solved in SBML.jl.

But I think there is an argument to be made that this expansion of `reateOf` should be handled by libSBML.

exaexa:
How would that extend to situations if there would be more reactions producing/consuming `p1`, depending with production rates on each other? (in short, this it would require SBML.jl to solve ODEs, which it won't).

OTOH the 'rateOf' should be handled somehow. Is there some reasonable julia representation of that? would `'` aka `prime` work?

exaexa:
(how's "derivation" marked in symbolics?)

paulflang:
AFK, but I don't think we need derivatives for this. We just need the sum of all the kineticLaws the species participates in. Times stoichiometry of course.

exaexa:
OK that's reinterpretation of the contents. Isn't there a function for that in libsbml already?

exaexa:
(side question, what if the kinetic law will depend on the output of the reaction you're writing the kinetic law for?)

paulflang:
Thanks for suggesting the cleaner way of reproducing the bug. To your question

[...]

PR #156, opened 2021-11-04
Add mapping of `rateOf` for Symbolics

giordano:
Ref #139.  @paulflang would this work for you?

### DIFF
--- a/src/symbolics.jl
+++ b/src/symbolics.jl
@@ -51,6 +51,7 @@ const default_symbolics_mapping = Dict{String,Any}(
     "or" => :|,
     "piecewise" => :(sbmlPiecewise),
     "power" => :^,
+    "rateOf" => :sbmlrateOf,
     "root" => :sbmlRoot,
     "sech" => :sech,
     "sec" => :sec,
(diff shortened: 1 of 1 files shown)


exaexa:
Looks great to me. Is there a SBML testcase that would trigger the code to get coverage?

paulflang:
I think we would have to automatically connect to the other reactions. Unless ModelingToolkit is very smart in its simplifications and does that for us. But it is definitely a step in the right direction.

giordano:
>  we would have to automatically connect to the other reactions.

Do we already do something like that elsewhere in `SBML.jl`?

paulflang:
`extensive_kinetic_math`
 does sth somewhat similar: looking up the species in a `kineticLaw` and depending on whether they have `hOSU` correcting for volume: https://github.com/LCSB-BioCore/SBML.jl/blob/1aa07d4fb21a361ae36cbb6963ad30640309f4eb/src/utils.jl#L193

exaexa:
Hey guys esp @paulflang is this okay to merge ie doesn't it kill any usecases for you? I'd smash it into 0.8.1 (reqd by anand in the other PR)

exaexa (approved):
after quick checking it doesn't seem to me this would break any package in the environs; I guess we're good to go

paulflang:
Don't think it kills anything, so should be fine.

PR #158, opened 2021-11-12
Fix typo in name of `sbmlRateOf` function

giordano:
Ref: https://github.com/LCSB-BioCore/SBML.jl/issues/139#issuecomment-967142733


exaexa:
maybe we should really have a test for this, can we add 01269 to the downloads and (maybe) check if rateOf is really there?

PR #160, opened 2021-11-15
Rollback (temporarily) the unitful units

exaexa:
...and add tests for the usecases that got broken. :]

This will need a slightly more systematic solution.


giordano (approved):

giordano:
For the benefit of future readers, the problem was that `flux_bounds` expects `Float64`:
https://github.com/LCSB-BioCore/SBML.jl/blob/0b3792f72bd30f103a021085716dce66898ab404/src/utils.jl#L55
but here we were returning `Tuple{Quantity,String}` (or more correctly `Tuple{Union{Float64,Quantity},String}`), and when there were unitful numbers, the automatic conversion to `Float64` was failing.

We're thinking about how to better serve the units to end users here.

Issue #134, opened 2021-07-26
case 00489 fails


paulflang:
## Minimal code example to reproduce the problem

```julia
using SBMLToolkit
using SBML

case = "00489"
url = "https://raw.githubusercontent.com/sbmlteam/sbml-test-suite/master/cases/semantic/$case/$case-sbml-l3v2.xml"
tmpfile = download(url)
try
    SBMLToolkit.checksupport(tmpfile)
    ml = readSBML(tmpfile, doc -> begin
            set_level_and_version(3, 2)(doc)
            convert_simplify_math(doc)
        end)
    ia = SBML.initial_amounts(ml, convert_concentrations = true)
    ia = [k=>v for (k,v) in ia]
    println(ia)
finally
    rm(tmpfile)
end
```

## Expected behavior

Should print:
```
Pair{String, B} where B["S1" => 2*p1, "S3" => 0.01, "S2" => 0.02]
```

## Actual behavior

Prints:
```
Pair{String, B} where B["S1" => nothing, "S3" => 0.01, "S2" => 0.02]
```

exaexa:
Hah, this needs some restructuralization in the code, because with this the compartment size (nor the concentrations it seems) are not definable by numbers.

Required changes:
- represent `initialAssignment`s
- `initial_amounts` would return `SBML.Math` instead of just Floats.

exaexa:
The original problem here needs a bit smarter evaluator (one that can reach for initial assignments automatically), but that should be completely and easily doable with #163 merged.

PR #163, opened 2021-11-19
Support initial assignments

giordano:
Very WIP, this will eventually address #134.  ~However, at the moment the model `00489` reported in that issue appears to have... no initial assignments, which is rather interesting.  Also, now we get a bunch of warnings about unsupported math, which cause a test failure.~

### DIFF
--- a/test/loadmodels.jl
+++ b/test/loadmodels.jl
@@ -81,6 +81,14 @@ sbmlfiles = [
         0,
         Float64[],
     ),
+    (
+        joinpath(@__DIR__, "data", "00489-sbml-l3v2.xml"),
+        "https://raw.githubusercontent.com/sbmlteam/sbml-test-suite/master/cases/semantic/00489/00489-sbml-l3v2.xml",
+        "dab2bce4e5036fa47ad8137055ca5f6dec6dfcb183542ce38573ca2e5a615813",
+        3,
+        2,
+        fill(Inf, 2),
+    ),
 ]
 
 @testset "Loading of models from various sources - $(reader)" for reader in (
--- a/test/loadmodels.jl
+++ b/test/loadmodels.jl
@@ -180,6 +180,19 @@ end
     )
 end
 
+@testset "Initial assignments" begin
+    m = readSBML(joinpath(@__DIR__, "data", "00489-sbml-l3v2.xml"))
+    @test m.initial_assignments == Dict(
+        "S1" => SBML.MathApply("*", [SBML.MathVal{Int32}(2), SBML.MathIdent("p1")]),
+    )
+
+    m = readSBML(joinpath(@__DIR__, "data", "sbml01289.xml"))
+    @test m.initial_assignments == Dict(
+        "p2" => SBML.MathApply("gt5", [SBML.MathVal{Int32}(8)]),
+        "p1" => SBML.MathApply("gt5", [SBML.MathVal{Int32}(3)]),
+    )
+end
+
 @testset "Extensive kinetic math" begin
     m = readSBML(joinpath(@__DIR__, "data", "sbml00852.xml"))
 
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -456,11 +456,23 @@ function extract_model(mdl::VPtr)::SBML.Model
             )
     end
 
+    initial_assignments = Dict{String,Math}()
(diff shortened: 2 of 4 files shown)


test/loadmodels.jl
```diff
@@ -214,7 +235,7 @@ end
     @test test_math.args[2].fn == "sin"
     @test test_math.args[2].args[1].val == 2.1
 
-    @test_logs (:warn,) (:warn,) (:warn,) (:warn,) readSBML(
+    @test_logs (:warn,) (:warn,) (:warn,) (:warn,) (:warn,) readSBML(
```
giordano, line 238: This is allowing an extra warning about an unsupported math element so that tests will pass for the time being, but we may want to fix that before merging this PR

exaexa (approved):
LGTM FFTM

Might be worth discussing the effect of math simplification with @paulflang . (Apparently, the initial assignments disappear in case the simplification is applied.)

paulflang:
Thanks, @giordano ! I am actually curious why we have #134 . I would have thought the libSBML `expandInitialAssignments` converter, which was called via the `expand_simplify_math` function should have taken care of this. But apparently it did not. So we needed a solution like this. I am just wondering if this solution follows the principle of least surprise. `initial_concentration` and `initial_amount` are attributes of `Species`, but `initial_assignment` an attribute of `Model`.

@exaexa : Yes, I think the `initial_amounts()` and `initial_concentrations()` functions would should eventually return floats and not mathematical expressions (at least optionally, or after having called some function that populates `Species.initial_amount` and `Species.initial_concentration` with the (un?)evaluated `initialAssignment`).

exaexa:
> I am actually curious why we have #134 

well you opened it :D

Anyway I guess it's solved now, you can get the initial assignments inside with expandInitialAssignments, and if that fails you can do that manually with any math evaluating function.

Notably, `Species_getInitialAmount` can _not_ return math, thus the `2*p1` case there isn't fixable by SBML, you need to reach manually to the `initialAmounts`. I guess we can close the case. :]

paulflang:
I know that I opened, it, but it still baffles me why the libSBML `expandInitialAssignments` converter has not handled it so that I had to open it.

Anyway. Do you think it is a good idea that `initial_assignment` is a field of `Model` rather than `Species`?

giordano:
> Anyway. Do you think it is a good idea that initial_assignment is a field of Model rather than Species?

Oh, yeah, that's a good question, I also wasn't sure `Model` was the best place.

paulflang:
> Apparently, the initial assignments disappear in case the simplification is applied.

Oh, I think I start to see what you mean @exaexa . The libSBML `expandInitialAssignments` converter does not copy the initialAssignments into initialAmounts, of species, but moves it there. And for whatever reason does not expand it either. But this is a problem, because like you said `Species_getInitialAmount` can not return math. So after running the libSBML `expandInitialAssignment` converter, the document is converted to a form that does no longer allow extraction of the initialAssignment/initialAmount using any libSBML functionality, right?
But if this were right, it would be a bug in libSBML, wouldn't it?

exaexa:
> Anyway. Do you think it is a good idea that initial_assignment is a field of Model rather than Species?

Yeah, SBML defines it like that. There might be initial assignments for other things as well.

> And for whatever reason does not expand it either.

Yeah, the reasoning behind that is completely dark to me. I can have a look at the source code and find either a bug or an unexpected explanation, but generally I'd really really recommend you guys to have a custom (partial) evaluator for the math. It is quite simple to do with Symbolics, and IMO will save you tons of trouble also with many other issues.

paulflang:
We cannot evaluate anything, cause the information is deleted after importing with `convert_simplify_math`, which we always do for plenty of reasons. Or do I see this wrong?

exaexa:
Ah so, I meant this as a kinda replacement for `simplify_math`.

paulflang:
I guess we could use some functions of `convert_simplify_math`, but not the `expandInitialAssignments` converter. We would have to write our own `expandInitialAssignments` converter that does the same thing as the libSBML converter, plus it handles the math evaluation. Alternatively, we could read in the model twice, once with the converter, once without the converter, and then overwrite initial conditions that are `nothing` with the evaluated math. But that would be a bit hacky. Shall I write an issue on the libSBML repo why the `expandInitialAssignments` converter does not expand the math?

PR #164, opened 2021-11-26
Add test cases for loading dynamic models and convert.jl

### LINKED ISSUE #56: Find and add good tests for loading converters

giordano:
Revive #64, possibly fix #56.  CC: @paulflang

### DIFF
--- a/test/loaddynamicmodels.jl
+++ b/test/loaddynamicmodels.jl
@@ -1,0 +1,59 @@
+using Test
+using SBML
+using Symbolics
+
+@testset "Loading of models from various sources" begin
+    @variables s1, s2, s3, s4, t, k1, c1, reaction_1_k1
+
+    sbmlfiles = [
+        (
+            joinpath(@__DIR__, "data", "reactionsystem_02.xml"),
+            Pair{String,Dict{String,String}}[],
+            1,
+            ("reaction", c1*k1*s1*s2)
+        ),
+        (
+            joinpath(@__DIR__, "data", "reactionsystem_02.xml"),
+            ["promoteLocalParameters" => Dict{String,String}()],
+            2,
+            ("reaction_1", c1*reaction_1_k1*s3)
+        ),
+        (
+            joinpath(@__DIR__, "data", "reactionsystem_02.xml"),
+            ["expandFunctionDefinitions" => Dict{String,String}()],
+            1,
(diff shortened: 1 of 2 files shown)


test/loaddynamicmodels.jl
```diff
@@ -0,0 +1,59 @@
...
+        ),
+        (
+            joinpath(@__DIR__, "data", "reactionsystem_03.xml"),
+            ["expandInitialAssignments" => Dict{String,String}()],
+            1,
+            ("reaction_2", c1*k1*s4)
+        ),
+        (
+            joinpath(@__DIR__, "data", "reactionsystem_04.xml"),
+            # PL: TODO: Unfortunately, libSBML has an inconsistent API here I think. Hence
```
exaexa, line 35: could we get a hint about what's "broken" if there's no level set?

giordano:
Ok, I had started this PR long time ago and it looks like I had also figured out where the input files were.  Tests are passing locally for me, but now I need to remember where I downloaded them from :grimacing:

giordano:
For my own record: the content of #64, before a force-push made its content completely disappear, was https://github.com/LCSB-BioCore/SBML.jl/commit/31acf728eeca28bf11b8ca9098e01f6426c8d088, which has somewhat different tests.

The tests I used here are from https://github.com/paulflang/SBML.jl/blob/pl/mk-species-units/test/loaddynamicmodels.jl, mentioned in #56.

test/loaddynamicmodels.jl
```diff
@@ -0,0 +1,91 @@
...
+            @test length(mdl.parameters) == expected_par
+
+            id, math = expected_rxn
+            @test isequal(repr(mdl.reactions[id].kinetic_math), repr(math))
+
+            id, ic = expected_u0
+            if basename(sbmlfile) == "00037-sbml-l3v2.xml"
+                # When expanding initial assignments with libsbml, the initial amount
+                # becomes empty.
+                @test_broken mdl.species[id].initial_amount[1] == ic
```
giordano, line 85: This is basically what we found in #163

exaexa (approved):
looks good, let's see. :D

PR #165, opened 2021-12-13
Events support

anandijain:
I'll need to add the delays and the `useValuesFromTriggerTime` attribute later, but this is a good starting point for what i need done in the short term. Since it doesn't seem that the delays can be used easily in MTK yet, I can put that off for another PR.

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -479,7 +479,7 @@ function extract_model(mdl::VPtr)::SBML.Model
             trig_math = parse_math(trig_math_ptr)
         end
     
-        event_assignments = []
+        event_assignments = EventAssignment[]
         num_event_assignments = ccall(sbml(:Event_getNumEventAssignments), Cuint, (VPtr,), ev)
         for j = 0:(num_event_assignments-1)
             eva = ccall(sbml(:Event_getEventAssignment), VPtr, (VPtr, Cuint), ev, j)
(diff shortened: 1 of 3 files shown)


exaexa (approved):
Looks good to me. We may wait for @giordano in case of any comments, but otherwise I think this is ready to merge.

src/structs.jl
```diff
@@ -196,6 +196,29 @@ end
 """
 $(TYPEDEF)
 
+# Fields
+$(TYPEDFIELDS)
+"""
+struct Event
+    name
```
exaexa, line 203: naaaah, this needs types.

exaexa (changes requested):

src/readsbml.jl
```diff
@@ -467,6 +467,35 @@ function extract_model(mdl::VPtr)::SBML.Model
...
+    for n = 0:(num_events-1)
+        ev = ccall(sbml(:Model_getEvent), VPtr, (VPtr, Cuint), mdl, n)
+        sym = ccall(sbml(:Event_getId), Cstring, (VPtr,), ev)
+        trig = ccall(sbml(:Event_getTrigger), VPtr, (VPtr,), ev)
+        trig_math_ptr = ccall(sbml(:Trigger_getMath), VPtr, (VPtr,), trig)
+        if trig_math_ptr != C_NULL
+            trig_math = parse_math(trig_math_ptr)
+        end
+    
+        event_assignments = []
```
giordano, line 482:
```suggestion
        event_assignments = EventAssignment[]
```

giordano (commented):
Looks good to me, apart from the missing type annotations

exaexa (approved):
cool thanks!

Issue #53, opened 2021-06-06
Rate rules 


anandijain:
> Rules are currently not imported into SBML.Model (and also not representable as ReactionSystem, I think). Should be imported into SBML.Model at some point to allow lowering to ODESystem.

cc @paulflang  

There are a big class of ODE SBML models that use these Rate laws, so it'd be good to get this in motion.

paulflang:
essentially what we need to import are the things in the SBML `listOfRules`.

exaexa:
Hi! As always, is there some example xml of that available publicly?

paulflang:
example: https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/00029/00029-sbml-l3v2.xml

If rules are implemented, we can get rid of this: https://github.com/paulflang/SBML.jl/blob/d714fb5014213744ba76f1a1b4cdb74b30689f8b/src/readsbml.jl#L178

paulflang:
I actually run into the issue that models with `Events` are not simulated correctly. I'd like to raise an not implemented error for that, but `Rules` are not represented in `SBML.Model`. Could you represent them? More generally, could you represent everything that is in `listOf___`. The only things that are still missing are (see p. 36 of the SBML specs)

- listOfInitialAssignments
- listOfRules
- listOfConstraints
- listOfEvents

paulflang:
Just to clarify, in my opinion SBML.Model should have a representation of all the `listOf___` attributes that are on [p36 of the specs](http://sbml.org/Special/specifications/sbml-level-3/version-2/core/release-2/sbml-level-3-version-2-release-2-core.pdf).

exaexa:
The `listOf*` is actually a complete antipattern, even in XML. Why would you want it?

paulflang:
To not simulate models incorrectly. I mean what is the antipattern here? We already have most of the listOf* stuff implemented:
- Compartments
- Species
- Reactions

Why not also implement Rules, Events, etc.?

For instance, ignoring the `Rules` in a model that contains them seems to be equivalent to ignoring some reactions at random in a model that contains only reactions (although, I've seen StackOverflow errors in models that contain rules, so at least they don't silently simulate incorrectly. But idk if that is the case for all models where we ignore the rules).

exaexa:
aaaaaaah you actually sounded like reimplementing the listOf* mechanics, that would be bad. Certainly we want all that stuff represented.

[...]

PR #166, opened 2021-12-14
Support list of rules

giordano:
Ref #53.

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -478,7 +478,7 @@ function extract_model(mdl::VPtr)::SBML.Model
         if trig_math_ptr != C_NULL
             trig_math = parse_math(trig_math_ptr)
         end
-    
+
         event_assignments = EventAssignment[]
         num_event_assignments = ccall(sbml(:Event_getNumEventAssignments), Cuint, (VPtr,), ev)
         for j = 0:(num_event_assignments-1)
@@ -495,6 +495,17 @@ function extract_model(mdl::VPtr)::SBML.Model
         events[evname] = SBML.Event(evname, trig_math, event_assignments)
     end
 
+    # Rules
+    rules = Dict{String,Math}()
+    num_rules = ccall(sbml(:Model_getNumRules), Cuint, (VPtr,), mdl)
+    for n in 0:(num_rules - 1)
+        rule = ccall(sbml(:Model_getRule), VPtr, (VPtr, Cuint), mdl, n)
+        var = ccall(sbml(:Rule_getVariable), Cstring, (VPtr,), rule)
+        math_ptr = ccall(sbml(:Rule_getMath), VPtr, (VPtr,), rule)
+        if math_ptr != C_NULL
+            rules[unsafe_string(var)] = parse_math(math_ptr)
+        end
+    end
 
     return Model(
         parameters,
@@ -502,6 +513,7 @@ function extract_model(mdl::VPtr)::SBML.Model
         compartments,
         species,
         initial_assignments,
+        rules,
         reactions,
         objective,
         gene_products,
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -496,14 +496,29 @@ function extract_model(mdl::VPtr)::SBML.Model
     end
 
     # Rules
-    rules = Dict{String,Math}()
+    rules = Rule[]
     num_rules = ccall(sbml(:Model_getNumRules), Cuint, (VPtr,), mdl)
     for n in 0:(num_rules - 1)
-        rule = ccall(sbml(:Model_getRule), VPtr, (VPtr, Cuint), mdl, n)
-        var = ccall(sbml(:Rule_getVariable), Cstring, (VPtr,), rule)
-        math_ptr = ccall(sbml(:Rule_getMath), VPtr, (VPtr,), rule)
+        rule_ptr = ccall(sbml(:Model_getRule), VPtr, (VPtr, Cuint), mdl, n)
+        type = if ccall(sbml(:Rule_isAlgebraic), Bool, (VPtr,), rule_ptr)
+            AlgebraicRule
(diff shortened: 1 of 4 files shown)


src/readsbml.jl
```diff
@@ -478,7 +478,7 @@ function extract_model(mdl::VPtr)::SBML.Model
         if trig_math_ptr != C_NULL
             trig_math = parse_math(trig_math_ptr)
         end
-    
+
```
exaexa, line 481: <3

src/readsbml.jl
```diff
@@ -495,13 +495,40 @@ function extract_model(mdl::VPtr)::SBML.Model
...
+        end
+    end
 
     return Model(
         parameters,
         units,
         compartments,
         species,
         initial_assignments,
+        rules,
```
exaexa, line 531: Not a concern here but I realized we should aim for extensibility and created #167

src/utils.jl
```diff
@@ -315,5 +315,8 @@ function Base.show(io::IO, ::MIME"text/plain", m::SBML.Model)
     )
 end
 
-# Define equality `==` method for `MathApply`.
+# Define equality `==` methods for some of our custom types.
 Base.:(==)(a::MathApply, b::MathApply) = a.fn == b.fn && a.args == b.args
+Base.:(==)(a::AlgebraicRule, b::AlgebraicRule) = a.math == b.math
+Base.:(==)(a::T, b::T) where {T<:Union{AssignmentRule,RateRule}} =
```
exaexa, line 321: This can compare AssignmentRule to be equal to RateRule right?

exaexa: (maybe not, I'm not sure if <: destroys the Union or not but I'd assume not.)

exaexa (commented):
looks great, 1 question to answer about the `==`

src/structs.jl
```diff
@@ -234,13 +279,14 @@ struct Model
     compartments::Dict{String,Compartment}
     species::Dict{String,Species}
     initial_assignments::Dict{String,Math}
+    rules::Vector{Rule}
```
giordano, line 282: This field is a bit of an outlier here: it's a vector, all others are either dictionaries or `Maybe{x}`.

exaexa: The rules do not have identifiers right? If not it's just a vector.

exaexa: :sob: tho.

src/utils.jl
giordano: No, they'd be different:
```julia
julia> SBML.RateRule("x", SBML.MathVal(1)) == SBML.AssignmentRule("x", SBML.MathVal(1))
false

julia> SBML.RateRule("x", SBML.MathVal(1)) == SBML.RateRule("x", SBML.MathVal(1))
true

julia> SBML.AssignmentRule("x", SBML.MathVal(1)) == SBML.AssignmentRule("x", SBML.MathVal(1))
true
```
A concrete object can have either `AssignmentRule` or `RateRule` type, so `T` would be only one type of the union and both arguments must have the same type.

src/structs.jl
giordano: Algebraic rules don't have an identifier: http://sbml.org/Software/libSBML/5.18.0/docs/formatted/c-api/class_rule__t.html.  And I think in principle you can have both assignment and rate rule for the same variable?  Not sure whether that's ruled out by the no-algebraic-loops rule.

exaexa: yes, I'd even say you can have 2 same assignments for the same variable, defining equality constraint. :D

But anyway yeah this vector is correct here. :]

src/utils.jl
exaexa: ok seems like T is forced to be rank1. :D Solved

exaexa (approved):
g r e a t

src/readsbml.jl
```diff
@@ -495,13 +495,40 @@ function extract_model(mdl::VPtr)::SBML.Model
...
+    num_rules = ccall(sbml(:Model_getNumRules), Cuint, (VPtr,), mdl)
+    for n in 0:(num_rules - 1)
+        rule_ptr = ccall(sbml(:Model_getRule), VPtr, (VPtr, Cuint), mdl, n)
+        type = if ccall(sbml(:Rule_isAlgebraic), Bool, (VPtr,), rule_ptr)
+            AlgebraicRule
+        elseif ccall(sbml(:Rule_isAssignment), Bool, (VPtr,), rule_ptr)
+            AssignmentRule
+        elseif ccall(sbml(:Rule_isRate), Bool, (VPtr,), rule_ptr)
+            RateRule
+        end
```
giordano, line 509: I wonder if this should have an `else` branch throwing an error, in case we get a weird (well, wrong) model with an unknown rule type?

exaexa: well not an error but can we have a warning about unsupported contents (or just ignore it because SBML says there's no more assignment types?)

giordano: Ok, well, at the moment `type` would be undefined if all branches are false, which is close to the "just ignore it" option.  I'm not even sure that's possible to have a rule not of one of those types, I suspect libsbml would skip it.

giordano: Yeah, if I change the name of one of the rules to `myRule` I get:
```
┌ Error: SBML reported error: Apart from the general Notes and Annotation subobjects permitted on all SBML components, a ListOfRules container object may only contain Rule objects.
│ Reference: L3V1 Section 4.2.8
│  Element 'myRule' is not part of the definition of <listOfRules>.
└ @ SBML ~/.julia/dev/SBML/src/utils.jl:271
ERROR: AssertionError: Opening SBML document has reported errors
```
So probably that isn't a concern in practice, libsbml would fail much earlier

exaexa: ok that sounds good. :]

PR #168, opened 2021-12-16
Define keywords-based constructors for `Model` and `Reaction`

### LINKED ISSUE #167: functions with more than ~3 parameters should be converted to keyword parameters
## Target functionality

position-checking the arguments in constructors of Model and Reaction isn't okay at all, we should do better :]

## Desired result

- we can have default values in the middle
- I won't mess the argument order anytime again
- extensibility!!!111 (adding functionality won't break everyone's code ever again)

giordano:
Fix #167, I think

### DIFF
--- a/src/structs.jl
+++ b/src/structs.jl
@@ -127,7 +127,7 @@ unit names), and objective coefficient (`oc`). Also may contains `notes` and
 # Fields
 $(TYPEDFIELDS)
 """
-struct Reaction
+Base.@kwdef struct Reaction
     reactants::Dict{String,Float64}
     products::Dict{String,Float64}
     kinetic_parameters::Dict{String,Tuple{Float64,String}}
@@ -136,10 +136,8 @@ struct Reaction
     gene_product_association::Maybe{GeneProductAssociation}
     kinetic_math::Maybe{Math}
     reversible::Bool
-    notes::Maybe{String}
-    annotation::Maybe{String}
-    Reaction(rs, prs, pas, l, u, as, km, r, n = nothing, an = nothing) =
-        new(rs, prs, pas, l, u, as, km, r, n, an)
+    notes::Maybe{String} = nothing
+    annotation::Maybe{String} = nothing
 end
 
 """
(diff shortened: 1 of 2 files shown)


giordano:
Tangentially related: the [docstring](https://lcsb-biocore.github.io/SBML.jl/dev/#SBML.Model) of `Model` reads
```
help?> SBML.Model
  struct Model

  Structure that collects the model-related data. Contains parameters, units, compartments, species and reactions and gene_products, and additional notes and annotation (also present internally in some of the data fields). The
  contained dictionaries are indexed by identifiers of the corresponding objects.

  Fields
  ≡≡≡≡≡≡≡≡

    •  parameters::Dict{String, Tuple{Float64, String}}
    •  units::Dict{String, Vector{SBML.UnitPart}}
    •  compartments::Dict{String, SBML.Compartment}
    •  species::Dict{String, SBML.Species}
    •  initial_assignments::Dict{String, SBML.Math}
    •  rules::Vector{SBML.Rule}
    •  reactions::Dict{String, SBML.Reaction}
    •  objective::Dict{String, Float64}
    •  gene_products::Dict{String, SBML.GeneProduct}
    •  function_definitions::Dict{String, SBML.FunctionDefinition}
    •  events::Dict{String, SBML.Event}
    •  notes::Union{Nothing, String}
    •  annotation::Union{Nothing, String}
```
Maybe we should remove "Contains parameters, units, compartments, species and reactions and gene_products, and additional notes and annotation (also present internally in some of the data fields)."?

exaexa (approved):
ok this is plain win

src/structs.jl
```diff
@@ -255,10 +248,10 @@ $(TYPEDEF)
 # Fields
 $(TYPEDFIELDS)
 """
-struct Event
-    name::Maybe{String}
-    trigger::Maybe{Math}
-    event_assignments::Vector{EventAssignment}
+Base.@kwdef struct Event
+    name::Maybe{String} = nothing
```
giordano, line 252: I'm not entirely sure whether you wanted to default this field to `nothing` or not

exaexa: yes, the names are "human IDs" and they can be missing by default. We can eventually add the nothings everywhere else too.

PR #169, opened 2021-12-16
name the fields in longer constructors

exaexa:
to avoid breaking everything everytime a field is added. :]

Also soaked up one related fix of the Event

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -472,28 +472,34 @@ function extract_model(mdl::VPtr)::SBML.Model
     num_events = ccall(sbml(:Model_getNumEvents), Cuint, (VPtr,), mdl)
     for n = 0:(num_events-1)
         ev = ccall(sbml(:Model_getEvent), VPtr, (VPtr, Cuint), mdl, n)
-        sym = ccall(sbml(:Event_getId), Cstring, (VPtr,), ev)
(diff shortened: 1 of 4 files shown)


giordano:
Ehm... #168 :stuck_out_tongue:

src/readsbml.jl
```diff
@@ -458,7 +458,7 @@ function extract_model(mdl::VPtr)::SBML.Model
 
     initial_assignments = Dict{String,Math}()
     num_ias = ccall(sbml(:Model_getNumInitialAssignments), Cuint, (VPtr,), mdl)
-    for n in 0:(num_ias - 1)
+    for n = 0:(num_ias-1)
```
giordano, line 461: Side note, the fact that you can do for loops with the `for var = range` syntax confuses some new users.  I wish that syntax wasn't allowed at all, and I'm not the only one.

exaexa: ha, this is actually enforced by juliaformatter. Should I complain there?

giordano: Ugh.  Yes, please, if you're willing to do it :smile:

exaexa: the point is that `in` collides nicely with the `in` operator in a way that is IMO even worse. Why's there even the separator when the parse is unambiguous anyway? :D

PR #170, opened 2021-12-20
Support list of constraints

giordano:
I believe this is the last `listOf*` listed in https://github.com/LCSB-BioCore/SBML.jl/issues/53#issuecomment-860864017 missing.

Note, I didn't find _any_ constraint example in the [`sbml-test-suite`](https://github.com/sbmlteam/sbml-test-suite), so I simply copied the example in SBML specification in our custom model.

### DIFF
--- a/test/loadmodels.jl
+++ b/test/loadmodels.jl
@@ -260,7 +260,7 @@ end
             SBML.MathVal{Float64}(100.0)
         ])
     ]),
-        "<message>\n  <p xmlns=\"http://www.w3.org/1999/xhtml\"> Species S1 is out of range. </p>\n</message>",
+        " Species S1 is out of range. ",
     )
 end
 
--- a/test/loadmodels.jl
+++ b/test/loadmodels.jl
@@ -247,6 +247,23 @@ end
     ]
 end
 
+@testset "Constraints" begin
+    m = readSBML(joinpath(@__DIR__, "data", "custom.xml"))
+    @test only(m.constraints) == SBML.Constraint(
+    SBML.MathApply("and", [
+        SBML.MathApply("lt", [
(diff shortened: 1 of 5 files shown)


test/loadmodels.jl
```diff
@@ -247,6 +247,23 @@ end
...
+        SBML.MathApply("lt", [
+            SBML.MathVal{Float64}(1.0),
+            SBML.MathIdent("S1")
+        ]),
+        SBML.MathApply("lt", [
+            SBML.MathIdent("S1"),
+            SBML.MathVal{Float64}(100.0)
+        ])
+    ]),
+        "<message>\n  <p xmlns=\"http://www.w3.org/1999/xhtml\"> Species S1 is out of range. </p>\n</message>",
```
giordano, line 263: I'm trying to play with the functions for the [`XMLNode_t`](http://sbml.org/Software/libSBML/5.18.0/docs/formatted/c-api/class_x_m_l_node__t.html) class but I still haven't figured out how to get only the message text, rather than the full XML node :thinking:

giordano: Ok, thanks to hints from my colleagues I found the incantation to extract only the text from an XML node.  It may not be 100% reliable, but should work for the time being.  `listOfConstraints` doesn't seem to be widely used, so it may not be worth spending more time on this until we have an actual use case.

exaexa (approved):
looks good to me, the `get_string_from_xmlnode` may be even useful elsewhere (likely in the future for CVTerms and maybe for annotations).
Thanks a lot

:christmas_tree:

exaexa:
Also I'll close #53 because I think this really finishes it, if extra features are needed we can use new feat request for these.

PR #172, opened 2021-12-22
Mention UCL contribution to the package

giordano:


exaexa:
Looks okay, can you pls add the unbreakable spaces before the logo to add the (required) padding?
image

(it looks like the spaces between files are 3 normal spaces, but they aren't. I didn't find a better way to smuggle this to markdown tho :(  )

exaexa:
@laurentheirendt I think there's no conflict with permedcoe&other acks here, but requesting review just to be triple sure. :D

laurentheirendt (approved):
I am ok with this, of course, but I let @weiguUL have a look as well 👍🏼

weiguUL:
> I am ok with this, of course, but I let @weiguUL have a look as well 👍🏼

I am fine. Could we use the full logo of PerMedCoE (to make all logos similar in size)? @exaexa

PR #173, opened 2021-12-27
supply sane defaults to most SBML structures

exaexa:
...so that people can actually omit the "empty" parameters, and enjoy the cool compatibility! :]


exaexa:
Version bump to 0.9.1 is added so that I can require this in COBREXA, hopefully ending the version re-bumping series :D

exaexa:
(are the tests failing because EBI is overloaded?)

laurentheirendt (approved):

PR #176, opened 2022-01-11
add Genie shield to README

exaexa:
Links here: https://pkgs.genieframework.com/?packages=SBML

The stats there kinda make me wonder why SBML downloads would peak so much around christmas eve. :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # SBML.jl
 
-| Build status | Documentation |
-|:---:|:---:|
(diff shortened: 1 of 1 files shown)


giordano (approved):
Over 1k, wow! :rocket:

anandijain:
does that 1k include ci?

giordano:
https://github.com/GenieFramework/PkgVizBoard/issues/2

PR #177, opened 2022-01-13
Add CompatHelper

ChrisRackauckas:


ChrisRackauckas:
it's old.

exaexa:
> it's old.

seems so.

exaexa:
I'm getting scared by the missing version check but yeah ok whatevs. :D

PR #178, opened 2022-01-14
Read `name`, `id` and `conversionFactor` model attributes

### LINKED ISSUE #175: Model attributes
Is there a way to get the rest of the attributes of the `<model metaid="_case00031" id="case00031" name="case00031" timeUnits="time">`. its not super important though, but its one of those things that could make the `show` nicer

giordano:
Fix #175.


src/readsbml.jl
```diff
@@ -576,6 +576,9 @@ function extract_model(mdl::VPtr)::SBML.Model
         gene_products,
         function_definitions,
         events,
+        name = get_optional_string(mdl, :Model_getName),
+        id = get_optional_string(mdl, :Model_getId),
+        conversion_factor = get_optional_string(mdl, :Model_getConversionFactor),
```
giordano, line 581: Maybe we should treat the conversion factor more extensively?  According to [SBML Level 3 Version 2 specification](http://co.mbine.org/specifications/sbml.level-3.version-2.core.release-2.pdf), section 4.2.6 "The `conversionFactor` attribute" (page 39)

> The attribute `conversionFactor` defines a global value inherited by all `Species` object instances that do not
define separate values for their `conversionFactor` attributes.

So the species should inherit it if they don't define their own.

exaexa (approved):
lovely, FFTM. The conversion factor logic doesn't need to be added IMO.

PR #180, opened 2022-01-17
aj/sbml base units fallback

anandijain:


test/data/.gitignore
```diff
@@ -1,3 +1,4 @@
 # Input files needed when running the tests will be downloaded here
 /*.xml
 !/custom.xml
+!/base_units.xml
```
giordano, line 4: Why you're ignoring this if you just added it?

src/unitful.jl
```diff
@@ -68,7 +68,13 @@ unitful(u::Vector{UnitPart}) = prod(unitful.(u))
 Computes a properly unitful value from a value-unit pair stored in the model
 `m`.
 """
-unitful(m::Model, val::Tuple{Float64,String}) = unitful(m.units[val[2]]) * val[1]
+function unitful(m::Model, val::Tuple{Float64,String})
+    if haskey(m.units, val[2])
+        unitful(m.units[val[2]]) * val[1]
+    else 
+        UNITFUL_KIND_STRING[val[2]] * val[1]
```
giordano, line 75: Is there anything in the specification about this?

anandijain: On the top of page 44 of the l3v2 spec, it says that the symbols you have in that dict are reserved, but I haven't read particularly closely

test/data/base_units.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sbml xmlns="http://www.sbml.org/sbml/level3/version2/core" level="3" version="2">
+    <model id="case00935" name="case00935">
```
exaexa, line 3: if this is a testcase from SBML, it should be taken from SBML. If not, we need to write down copyright.

test/data/base_units.xml
```diff
@@ -0,0 +1,47 @@
...
+                            <ci> compartment </ci>
+                            <ci> k1 </ci>
+                            <ci> S1 </ci>
+                        </apply>
+                    </math>
+                </kineticLaw>
+            </reaction>
+        </listOfReactions>
+    </model>
+</sbml>
```
exaexa, line 47: final LF

src/unitful.jl
exaexa: the defaulting here probably needs to be a bit more complex. It should be able to reasonably default to e.g. substance_units for substance-ish stuff. That needs a slight modification of the API because the `unitful` function can't see where the Tuple comes from (if it is a volume or flux or what).

I'd suggest going with something like
```julia
unitful(m,  (0.1,""), :volume)
```
that defaults to 0.1 of the default volume units.

test/loadmodels.jl
```diff
@@ -192,6 +192,9 @@ end
     m = readSBML(joinpath(@__DIR__, "data", "custom.xml"))
     @test SBML.unitful(m.units["non_existent"]) == 0.00314
     @test SBML.unitful(m.units["no_dimensions"]) == 20.0
+
+    m = readSBML(joinpath(@__DIR__, "data", "base_units.xml"))
+    !haskey(m.units, "mole") && (SBML.unitful(m, (1., "mole")) == 1 * u"mol")
```
exaexa, line 197: this should be wrapped in `@test` right?

exaexa (changes requested):

src/unitful.jl
exaexa: Anyway, I didn't find any notice of this kind of defaulting in the SBML docs (the closest "reserved" word relates to base units). The spec is quite unhelpful on specifying on how to interpret the data.

Can you please point out the precise sections&paragraphs that warrant that this defaulting is right?

exaexa (changes requested):

src/unitful.jl
exaexa: whoa, is the model online? (does it pass the SBML validator? :D )

test/data/base_units.xml
giordano: It definitely doesn't look like https://github.com/sbmlteam/sbml-test-suite/blob/42cbda7a271053eccb90034e42d5c95a6cb049ce/cases/semantic/00935/00935-sbml-l3v2.xml

PR #181, opened 2022-01-26
make parameter a real struct

anandijain:
@exaexa Locally all tests are passing except something in flux_bounds fails. It should be an obvious change.
The reason I need this is I have a model with units in params.

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -147,7 +147,7 @@ end)
 function readSBML(
     fn::String,
     sbml_conversion = document -> nothing;
-    report_severities = ["Fatal", "Error"],
+    report_severities = ["Fatal", "Error"]
 )::SBML.Model
     isfile(fn) || throw(AssertionError("$(fn) is not a file"))
     _readSBML(:readSBML, fn, sbml_conversion, report_severities)
@@ -167,7 +167,7 @@ used to read from a file instead of a string.
 readSBMLFromString(
     str::AbstractString,
     sbml_conversion = document -> nothing;
-    report_severities = ["Fatal", "Error"],
+    report_severities = ["Fatal", "Error"]
 )::SBML.Model =
     _readSBML(:readSBMLFromString, String(str), sbml_conversion, report_severities)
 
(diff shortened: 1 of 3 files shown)


src/readsbml.jl
```diff
@@ -147,7 +147,7 @@ end)
 function readSBML(
     fn::String,
     sbml_conversion = document -> nothing;
-    report_severities = ["Fatal", "Error"],
+    report_severities = ["Fatal", "Error"]
```
giordano, line 150: Why removing the commas?  They're useful to add a new argument without disrupting git history, which is exactly what you're doing here.

anandijain: i think the vscode auto formatter did that

anandijain:
I will add a test, but unfortunately, I can't find a single sbml-test-suite case that actually uses units in parameters. 
I grepped with  `grep -R -l "<parameter.*unit"` and got nothing. Mose, can you sanity check before I open an issue in the test-suite repo. I feel like they should probably have at least one.

src/readsbml.jl
exaexa: JuliaFormatter to the git hook! :D

exaexa:
This kinda materializes `Parameter_t`, right?

Looks good to me, except we might stay consistent with the rest of SBML.jl and report units in normal tuples. (not that I'd like to support the technical debt in that part, but it's going to make the transition much easier)

anandijain:
I'm not exactly sure what you mean wrt the last part of tuples. What's the best way to go about that?
I could define `convert(Tuple, Parameter)` right? seems weird

exaexa:
> I'm not exactly sure what you mean wrt the last part of tuples. What's the best way to go about that?

The rest of SBML.jl represents all values in tuples of `(x::Real,unit::String)`; if you maintain that convention then both working with the current code _and_ the later migration to something better will be easier.

anandijain:
I see, what is a good way to do that though? This latest commit demonstrates my confusion haha

anandijain:
okay this should pass but I'm sure you have a way to clean it up

exaexa:
I have added a few commits, will merge this branch as a part of #182 . Thanks! (also feel free to review there :D )

PR #182, opened 2022-01-27
Clean up the parameter handling

exaexa:
This includes AJ's commits from #181 which is thus deprecated. Parameters are thus materialized with constantness status, names and proper units. Breaks compatibility with software that creates the old Parameter tuples directly (COBREXA).

Also closes #181.

### DIFF
--- a/src/structs.jl
+++ b/src/structs.jl
@@ -100,6 +100,20 @@ end
 """
 $(TYPEDEF)
 
+# Fields
+$(TYPEDFIELDS)
+"""
+Base.@kwdef struct Parameter
+    # id::String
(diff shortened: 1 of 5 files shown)


src/structs.jl
```diff
@@ -100,6 +100,19 @@ end
 """
 $(TYPEDEF)
```
giordano, line 102: Maybe add a short description of the type, besides the automatic stuff?

giordano (approved):

src/structs.jl
exaexa: very true, added below

anandijain:
thanks a bunch mirek!

PR #183, opened 2022-01-27
separate substance units from initial amounts and concentrations

exaexa:
SBML has the substance units only once, in `Species_t`.

Code was developed atop #182 so I base it on that branch.


giordano (approved):
Always love changes with negative net diff (`+16/-29`) :grin:

exaexa:
> Always love changes with negative net diff (`+16/-29`) grin

Yeah, the world has already enough software lines, line-adding diffs should be banned! :D

PR #184, opened 2022-02-02
extract reaction names

exaexa:
~~also bump the version for others to depend on~~
EDIT: not now, there's another breaking change


exaexa:
re :x: -- COBREXA compatibility is fixed in https://github.com/LCSB-BioCore/COBREXA.jl/pull/568

giordano (approved):

anandijain:
can i get constant attribute for species?

PR #187, opened 2022-02-24
make constant attribute mandatory for compartments, species, params

anandijain:
cf. l3v2 spec sections 4.5-4.7.

also boundaryCondition and hasOnlySubstanceUnit are mandatory for species


exaexa:
Is it possible to provide a default there? (so that it breaks the other packages a bit less)

exaexa:
(`constant` seems to have defaulted to `true` before)

exaexa:
btw is there a reason for breaking sbml less than L3V2 ?

anandijain:
> Is it possible to provide a default there? (so that it breaks the other packages a bit less)

it doesn't seem to mention anything about defaults, so I'd be hesitant to add them. If we need to, the sanest defaults are that species isn't constant, and params/compartments are constant.


> btw is there a reason for breaking sbml less than L3V2 ?

can you expand on this?

anandijain:
but yeah I realize this is sort of annoying to keep doing these breaking things

exaexa:
> can you expand on this? I'd hope that the converter that sets level/version would hopefully still work.

This way you're forcing people to reinterpret the contents of older models, possibly misinterpreting them/losing information.

(Also note how SBML spec actually handles this -- a missing value is technically possible, can be saved, but the file does not pass semantic validation.)

> but yeah I realize this is sort of annoying to keep doing these breaking things

Providing a default would work I guess. What does SBML fill in when it's converting from say L1 to L3 if `constant` is missing?

Still, this change actually _restricts_ the set of the models we can open, and does not add any actual value on the output apart from _danger_ of misinterpretation. Does it help you with something specific?

PR #188, opened 2022-03-25
Avoid using `eval`

giordano:
NOTE: this requires the functions in `mapping` to be accessible from `SBML`.  This is true for the functions in `default_symbolics_mapping`, but if the users passes something else, that's going to go boom.

CC: @anandijain


exaexa:
closing in favor of slightly more brutally functional solution in #189, but thanks for quickfixing the issue!

PR #189, opened 2022-03-25
avoid eval by directly using functions

exaexa:
alternative for #188

### DIFF
--- a/src/symbolics.jl
+++ b/src/symbolics.jl
@@ -22,7 +22,7 @@ sbmlRoot(x) = sqrt(x)
 sbmlRoot(power, x) = x^(1 / power)
 
 """
-    default_symbolics_mapping :: Dict{String,Function}
+    default_symbolics_mapping :: Dict{String,Any}
 
 Default mapping of SBML function names to Julia functions, represented as a
 dictionary from Strings (SBML names) to functions.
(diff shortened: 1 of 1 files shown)


giordano (approved):
As I said on Slack, if you don't really need symbols for a specific reason, proper functions are definitely better :+1:

src/symbolics.jl
```diff
@@ -82,6 +21,68 @@ sbmlLog(base, x) = log(base, x)
...
+    default_symbolics_mapping :: Dict{String,Function}
+
+Default mapping of SBML function names to Julia functions, represented as a
+dictionary from Strings (SBML names) to functions.
+
+The default mapping only contains the basic SBML functions that are
+unambiguously represented in Julia; it is supposed to be extended by the user
+if more functions need to be supported.
+"""
+const default_symbolics_mapping = Dict{String,Function}(
```
giordano, line 34: Actually, one minor nitpick: this could be
```suggestion
const default_symbolics_mapping = Dict{String,Any}(
```
`Function` is abstract, so it isn't any better than `Any`, really, and not all "callable" objects in Julia are functions (think of type constructors or Python functions from `PyCall`, for example)

exaexa:
Note to self: this found `not` actually not being defined.

src/symbolics.jl
exaexa: ok that's good to know... is there any supertype for callable stuff? if not I'm okay with Any

PR #191, opened 2022-05-18
allow dynamic compartment sizes

paulflang:


paulflang:
Fixes simulation of test suite case 00140, where the compartment volume is overwritten by an AssignmentRule.

src/utils.jl
```diff
@@ -241,9 +241,14 @@ function extensive_kinetic_math(
         haskey(m.species, x.id) || return x
         sp = m.species[x.id]
         sp.only_substance_units && return x
-        sz = m.compartments[sp.compartment].size
+        
+        cn = sp.compartment
+        compartment = m.compartments[cn]
+        rn = [r.id for r in m.rules if r isa SBML.AssignmentRule]
+        idx = findfirst(x->x==cn, rn)
+        sz = isnothing(idx) ? SBML.MathVal(compartment.size) : m.rules[idx].math
```
exaexa, line 249:
```suggestion
        # take the size from assignment rule math
        sz = maylift(i -> m.rules[i].math, findfirst(x -> (x isa AssignmentRule && x.id == sp.compartment), m.rules))
        # if there is no assignment, take the size from compartment structure
        isnothing(sz) && sz=SBML.MathVal(compartment.size)
        # otherwise fallback to user resolution
```

(not really tested tho)

exaexa: (btw I just noticed -- the default for handle_empty_compartment_size in documentation should return `SBML.MathVal` as well.)

exaexa (changes requested):
- Needs a test
- function documentation must be updated
- the line of fallbacks in the `conv()` function should be made clear, see suggestion

exaexa:
Deprecated by #194

PR #194, opened 2022-05-19
use compartment id instead of size in extensive_kinetic_math()

paulflang:
Still needs testing.


exaexa:
This deprecates #191 right?

paulflang:
yes, if i get it to work.

exaexa:
I rebased it to current master but can't push here, opening #199 to make the merge quick&easy :]

Please check it there, I cleaned some of the functions to be slightly more universal/readable and utilize the new `interpret_math`.

PR #196, opened 2022-05-19
Read also name of unit definitions

giordano:
This is an optional property and it's usually the same as the ID, reason why I
initially skipped it, but it may be useful to store this information as well for
consistency with the SBML format.

I noticed this while working on #143: by comparing the original file and the one we write there was this property missing.

Note: at the moment this is missing docstrings, will add them if this is deemed useful.
Note 2: this is a potentially breaking change as it modifies the layout of `Model`.


giordano:
Oh well, downstream tests didn't fail, at least this shouldn't break _too_ much code  :slightly_smiling_face:

exaexa (approved):
Looks good, if possible change the `list_of_units` to more correct `unit_parts`. (technically it's not even a list, even in libsbml :D)

PR #198, opened 2022-05-23
externalize the support for converting to Symbolics

### LINKED ISSUE #195: Remove direct dependency on Symbolics
## Target functionality

Symbolics load for ~7s for me because of precompilations, and the time goes up when using other numeric packages. In distributed processing this really hurts.

## Desired output

It is possible to parametrize out the total 3 references to `Symboilics.*` in `symbolics.jl`. The usecases will still be supported, users will have to add ~3 simple parameters to their code (number wrapper, variable wrapper, and differential mark). Properly documented this should add much performance and cause minimal breakage. Scheduling for 1.0.

exaexa:
This closes #195, and is thus breaking.

Users who do the `convert(Num, ...)` will need to copypaste a tiny bit of code that actually glues the new math interpreter to `Symbolics` -- stuff should work pretty much as-is with the code from `test/symbolics.jl`, you only need to rename the `interpret_as_num(...` to `convert(::Type{Num}, ...`.

I actually quite like this new version, because people can now rewrite SBML stuff pretty easily to e.g. expressions, throwing them to eg. ForwardDiff or just recreating SBML stuff as Julia functions.

The "default" functionality in `interpret` is just that of normal evaluation.


paulflang (approved):

src/interpret.jl
```diff
@@ -0,0 +1,139 @@
...
+    elseif length(args) >= 3
+        IfElse.ifelse(args[2], args[1], sbmlPiecewise(args[3:end]...))
+    else
+        throw(DomainError(args, "malformed piecewise SBML function"))
+    end
+end
+
+sbmlNeq(a, b) = !isequal(a, b)
+
+sbmlLog(x) = log(x, 10)
```
giordano, line 16:
```suggestion
sbmlLog(x) = log10(x)
```

test/runtests.jl
```diff
@@ -4,10 +4,12 @@ using SBML
 using SBML: Model, Reaction, Species
 using Symbolics, Unitful
 
+include("symbolics.jl")
```
giordano, line 7: Why outside of the rest of the test suite?

test/runtests.jl
```diff
@@ -4,10 +4,12 @@ using SBML
 using SBML: Model, Reaction, Species
 using Symbolics, Unitful
```
giordano, line 5: Can we move `using Symbolics` inside the `symbolics.jl` file?  The rationale is that if I want to do quick tests I can simply comment out the `include` line :slightly_smiling_face:

giordano (approved):
Looks good, as far as I'm concerned :rocket:

test/runtests.jl
exaexa: there are extra tests that depend on this. I'll make a condition there so that it can be switched off easily.

test/runtests.jl
exaexa: it's a library, not a testset

exaexa: ok I guess I solved this. :D

PR #199, opened 2022-05-24
extensive math correctly exposes compartment identifications

exaexa:
(originally #194, but rebase failed there and I'm quickfixing it here)

cc @paulflang

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -216,7 +216,7 @@ Recursively determine if an SBML.Math expression occurs in another.
 """
 function Base.occursin(needle::SBML.Math, haystack::SBML.Math)
     if haystack isa Union{SBML.MathApply, SBML.MathLambda}
-        return any(occursin.(needle, haystack.args))
+        return any([occursin(needle, arg) for arg in haystack.args])
     else
         needle == haystack && return true
     end
@@ -233,8 +233,8 @@ Determine if the compartmint with id `c_id` has a size.
 """
 function hassize(c_id::String, m::SBML.Model)
     isnothing(m.compartments[c_id].size) || return true
-    c_id in [r.id for r in m.rules] && return true
-    any(occursin.(SBML.MathIdent(c_id), [r.math for r in m.rules if r isa SBML.AlgebraicRule])) && return true
+    c_id in [r.id for r in m.rules if r isa Union{SBML.AssignmentRule, SBML.RateRule}] && return true
+    any([occursin(SBML.MathIdent(c_id), r.math) for r in m.rules if r isa SBML.AlgebraicRule]) && return true
     false
 end
 
@@ -274,8 +274,9 @@ function extensive_kinetic_math(
         haskey(m.species, x.id) || return x
         sp = m.species[x.id]
         sp.only_substance_units && return x
-        m.compartments[sp.compartment].spatial_dimensions == 0 && return x
-        hassize(sp.compartment, m) || handle_empty_compartment_size(x.id)
+        c = m.compartments[sp.compartment]
+        hassize(sp.compartment, m) || c.spatial_dimensions == 0 || handle_empty_compartment_size(x.id)
+        c.spatial_dimensions == 0 && return x
         SBML.MathApply("/", [x, SBML.MathIdent(sp.compartment)])
     end
     conv(x::SBML.MathApply) = SBML.MathApply(x.fn, conv.(x.args))
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -257,8 +257,7 @@ function extensive_kinetic_math(
         haskey(m.species, x.id) || return x
         sp = m.species[x.id]
         sp.only_substance_units && return x
-        # sz = m.compartments[sp.compartment].size
-        # isnothing(sz) && (sz = handle_empty_compartment_size(x.id))
+        m.compartments[sp.compartment].spatial_dimensions == 0 && return x
         hassize(sp.compartment, m) || handle_empty_compartment_size(x.id)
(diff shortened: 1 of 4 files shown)


src/utils.jl
```diff
@@ -205,6 +211,32 @@ initial_concentrations(
...
 )
 
+"""
+    isfreein(id::String, expr::SBML.Math)
+
+Determine if `id` is used and not bound (aka. free) in `expr`.
+"""
+isfreein(id::String, expr::SBML.Math) = interpret_math(
+    expr,
+    map_apply = any(rec.(x.body)),
```
paulflang, line 221: you sure this should not be `map_apply = x -> any(rec.(x.args...)),`?

src/utils.jl
```diff
@@ -205,6 +211,32 @@ initial_concentrations(
...
+    isboundbyrules(
+        id::String,
+        m::SBML.Model
+    )
+
+Determine if an identifier is defined or used by any Rules in the model.
+"""
+isboundbyrules(id::String, m::SBML.Model) =
+    any(r.id == id for r in m.rules if typeof(r) in [SBML.AssignmentRule, SBML.RateRule]) ||
+    any(isfreein(Ref(id), r.math for r in m.rules))
```
paulflang, line 239: I thought this needs to be `any(isfreein(Ref(id), r.math for r in m.rules if r isa SBML.AlgebraicRule))`. Cause if say a compartment appears on the rhs of an Assignment or RateRule it is not bound by that rule. Only if it appears on the lhs it is (which is exactly what is checked in the line above).

src/utils.jl
```diff
@@ -228,28 +260,35 @@ whatever units), you can specify that behavior by supplying
...
-        SBML.MathApply("/", [x, SBML.MathVal(sz)])
-    end
-    conv(x::SBML.MathApply) = SBML.MathApply(x.fn, conv.(x.args))
-    conv(x::SBML.Math) = x
-
-    conv(formula)
-end
+        isnothing(sz) &&
+            !isboundbyrules(sp.compartment, m) &&
+            (sz = handle_empty_compartment_size(x.id))
```
paulflang, line 282: In a case where the size is nothing and the compartment is not bound by rule we end up throwing an error here. However, before we do that, we should check whether `c.spatial_dimensions == 0`, in which case we set `sz = 1.` and don't need to throw an error, right?

src/utils.jl
```diff
@@ -228,28 +260,35 @@ whatever units), you can specify that behavior by supplying
...
+        isnothing(sz) && c.spatial_dimensions == 0 && (sz = 1.0)
```
paulflang, line 283: `sz` is not further used. IIUC we want to return the species without doing any division, i.e.: `isnothing(sz) && c.spatial_dimensions == 0 && return x`

test/loadmodels.jl
```diff
@@ -286,7 +295,14 @@ end
...
     @test subterm.args[1] == SBML.MathIdent("S1")
-    @test isapprox(subterm.args[2].val, 1.0)
+    @test subterm.args[2] == SBML.MathVal(1.0)
+
+    m = readSBML(joinpath(@__DIR__, "data", "sbml00140.xml"))
+
+    subterm = SBML.extensive_kinetic_math(m, m.reactions["reaction1"].kinetic_math).args[2]
+    @test subterm.fn == "/"
+    @test subterm.args[1] == SBML.MathIdent("S1")
+    @test subterm.args[2] == SBML.MathVal(5.0)
```
exaexa, line 305: @paulflang Now I'm at this -- the size gets substitued in there if it is available in compartment.size. Is this still desirable? If people are supposed to substitute for compartment variables themselves, I can remove the functionality altogether.

src/utils.jl
paulflang: I see @exaexa . I mean we never divide by a Float in extensive_kinetic_math. Either we divide by compartment.size or we return `species123` without division, which obviously is the same as dividing by 1.0.
But we only ignore compartment size if `isnothing(compartment.size) && !(compartment.spatial_dimensions == 0)`. However, in these cases `rateof(somethingElse) = somefunction ( compartment123 )` would have to use a `nothing` sized compartment and the whole simulation would fail there. And tbh, I think it should. A `nothing` sized compartment should not appear on the rhs of a Rate or AssignmentRule.

exaexa: Ah yes that probably corresponds to my other question below, I'd say it's cleared now if what I assumed there is right

src/utils.jl
```diff
@@ -228,28 +256,31 @@ whatever units), you can specify that behavior by supplying
...
-    end
-    conv(x::SBML.MathApply) = SBML.MathApply(x.fn, conv.(x.args))
-    conv(x::SBML.Math) = x
-
-    conv(formula)
-end
+        sz =
+            if !isboundbyrules(sp.compartment, m) &&
+               m.compartments[sp.compartment].spatial_dimensions == 0
+                # if the comparment ID isn't bound anywhere and it is a zero-dimensional compartment, default to 1.0
```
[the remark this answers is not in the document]
paulflang: > In addition to that, I still think that isboundbyrules should be

The reason why I think that is

- If a compartment appears on the rhs of a Assignment or RateRule, this does not bind its volume to anything. It's volume may well stay nothing. In that case, when the simulator uses the compartment id, we will get sth like `MethodError: no method matching /(::Float64, ::Nothing)`. But if the compartment size were zero dimensional, I would like to `sz = SBML.MathVal(1.0)  # or even && return x`.
- The problem you described (erroneously erased coupling) would not remain undetected. During simulation, we would get something like `MethodError: no method matching somefunction(::Nothing)`.

paulflang (approved):

PR #200, opened 2022-05-26
DocStringExtensions + some extra fixes

exaexa:
- `interpret_math` fix backported from #199 
- convert everything to DSE 0.9 because that can wrap long function definitions! (see https://github.com/JuliaDocs/DocStringExtensions.jl/pull/128# )

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -11,11 +11,12 @@ makedocs(
     authors = "The developers of SBML.jl",
     linkcheck = !("skiplinks" in ARGS),
     pages = ["Home" => "index.md", "Reference" => "functions.md"],
+    strict = [:missing_docs, :cross_references],
 )
 
 deploydocs(
     repo = "github.com/LCSB-BioCore/SBML.jl.git",
     target = "build",
     branch = "gh-pages",
-    push_preview = true,
+    push_preview = false,
 )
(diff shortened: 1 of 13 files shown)


docs/make.jl
```diff
@@ -11,11 +11,12 @@ makedocs(
...
     pages = ["Home" => "index.md", "Reference" => "functions.md"],
+    strict = [:missing_docs, :cross_references],
 )
 
 deploydocs(
     repo = "github.com/LCSB-BioCore/SBML.jl.git",
     target = "build",
     branch = "gh-pages",
-    push_preview = true,
+    push_preview = false,
```
giordano, line 21: Out of curiosity, why?

giordano (approved):

docs/make.jl
exaexa: I guess we don't really need the PR doc to be deployed.

Not that it wouldn't be useful at times, but for most occasions the check from PR is perfectly sufficient, and there's no reason to keep the preview docs rotting in gh-pages forever :D

giordano: You can automatically delete the directory after the PR is closed, see for example https://github.com/JuliaParallel/MPI.jl/blob/f8e4749c1ce8141458d0a88664044ec64da357ba/.github/workflows/PreviewCleanup.yml

exaexa: uuh cool. Writing that down. :D

PR #203, opened 2022-05-28
further formatter improvements

exaexa:
Github actions battle continues!

this 
- prevents ordinary issue comments from triggering the action (the `if` is global for the whole job)
- detect if github_token secret is available and attempts to fail gracefully if not (producing a CI fail instead of friendly warning comment)


exaexa (commented):
I forgot that this existed and just re-read it, taking that as a valid form of self-review.

Also not really sure I'll spot more bugs until this is really merged. :D

PR #204, opened 2022-06-04
fix isfreein()

paulflang:
I think this still needs a test case, but Idk where to put it best. For me, this failed with test suite case 39, which uses an algebraicRule with math `SBML.MathApply("+", SBML.Math[SBML.MathApply("+", SBML.Math[SBML.MathApply("*", SBML.Math[SBML.MathVal{Int32}(-1), SBML.MathIdent("k1")]), SBML.MathIdent("S1")]), SBML.MathIdent("S2")])`.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -210,7 +210,7 @@ Determine if `id` is used and not bound (aka. free) in `expr`.
 """
 isfreein(id::String, expr::SBML.Math) = interpret_math(
     expr,
-    map_apply = any(rec.(x.args)),
+    map_apply = (x, rec) -> any(rec.(x.args)),
     map_const = _ -> false,
     map_ident = x -> x.id == id,
     map_lambda = (x, rec) -> id in x.args ? false : rec(x.body),


exaexa:
Seems correct. Can you point out a test case for this? (Lack thereof is likely to blame here. :D)

If I get it right, this can be tested just by evaluating the math in sbml test 00039 ?

src/utils.jl
```diff
@@ -210,7 +210,7 @@ Determine if `id` is used and not bound (aka. free) in `expr`.
 """
 isfreein(id::String, expr::SBML.Math) = interpret_math(
     expr,
-    map_apply = any(rec.(x.args)),
+    map_apply = (x, rec) -> any(rec.(x.args)),
```
exaexa, line 213: This bug is me spoiled on haskells. I wish we had function types and a typechecker. :cry:

exaexa (approved):
thanks!

src/utils.jl
paulflang: > If I get it right, this can be tested just by evaluating the math in sbml test 00039

Yes.

PR #205, opened 2022-06-10
Test starting from Julia v1.6 in CI

giordano:
`SBML.jl` already requires Julia v1.6: https://github.com/LCSB-BioCore/SBML.jl/blob/6a124f994be04f33f2398f00b8676e1c258c3546/Project.toml#L19 but we're currently testing Julia v1.5.


exaexa (approved):
very true, thanks!

PR #210, opened 2022-06-14
do not store kinetic law parameters as global parameters

exaexa:
(this was probably introduced by mistake)


giordano:
Any meaningful test we can add?

paulflang:
AFAIK the parameters stored in kinetic laws are local to that kinetic law.

paulflang:
Anyway, looks good to me. But please up SBMLTk to get the downstream test working and make sure I haven't missed anything.

exaexa:
(Scheduled after WriteSBML is merged)

PR #211, opened 2022-06-17
Add `Objective` struct

giordano:
The `Objective_t` class has two required attributes which we're currently
completely missing: id and type.

Opening as draft because I'm not really sure how to deal with the
`flux_objective` function.

This is a breaking change.


giordano:
Note: we also need to mark the active objective: https://sbml.org/software/libsbml/5.18.0/docs/formatted/c-api/class_objective__t.html. This is especially important for writing the model to file (#143).

exaexa:
> Note: we also need to mark the active objective

uh yes that would be great (would simplify the failure logic I have in the PR to your branch)

exaexa (approved):
beautiful.

(Please merge when formatting is applied)

Issue #213, opened 2022-07-07
Add `exponentiale` to `default_constants`


paulflang:
https://github.com/LCSB-BioCore/SBML.jl/blob/b89c973c3da2f4cfd115a128fb2d4d33ea10c84e/src/interpret.jl#L97

Case 892 needs this. Page 21 in the specs describes this.

paulflang:
Yes, this is also how I get it.

exaexa:
Closed by #215

PR #215, opened 2022-07-13
Pl/interpret fixes

### LINKED ISSUE #214: Power fails with negative exponents
https://github.com/LCSB-BioCore/SBML.jl/blob/b89c973c3da2f4cfd115a128fb2d4d33ea10c84e/src/interpret.jl#L74

In case 889 I get:

```
ERROR: DomainError with -1:
Cannot raise an integer x to a negative power -1.
Make x or -1 a float by adding a zero decimal (e.g., 2.0^-1 or 2^-1.0 instead of 2^-1), or write 1/x^1, float(x)^-1, x^float(-1) or (x//1)^-1
Stacktrace:

throw_domerr_powbysq(#unused#::Int32, p::Int32) at [.\intfuncs.jl](vscode-file://vscode-app/c:/Users/wolf5212/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

power_by_squaring(x_::Int32, p::Int32) at [.\intfuncs.jl](vscode-file://vscode-app/c:/Users/wolf5212/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

^(x::Int32, p::Int32) at [.\intfuncs.jl](vscode-file://vscode-app/c:/Users/wolf5212/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

^(a::Num, b::Num) at [C:\Users\wolf5212.julia\packages\SymbolicUtils\9IZDP\src\methods.jl](vscode-file://vscode-app/c:/Users/wolf5212/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

(::SBMLToolkit.var"#17#22")(x::SBML.MathApply, interpret::Function) at ...
```

paulflang:
This PR shall fix #214  and #213 . Unfortunately, test fails for fixing #214 , but I still hope this gives a rough idea of what is needed.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -16,6 +16,7 @@ sbmlNeq(a, b) = !isequal(a, b)
 sbmlLog(x) = sbmlLog(10, x)
 sbmlLog(base, x) = log(base, x)
 
+sbmlPower(x, y) = IfElse.ifelse(y < 0, x^float(y), x^y)  # Todo @mirek: perhaps this needs a method float(y::SBML.MathVal)?
 sbmlRoot(x) = sqrt(x)
 sbmlRoot(power, x) = x^(1 / power)
 
@@ -71,7 +72,7 @@ const default_function_mapping = Dict{String,Any}(
     "not" => !,
     "or" => |,
     "piecewise" => sbmlPiecewise,
-    "power" => ^,
+    "power" => sbmlPower,
     "rateOf" => sbmlRateOf,
     "root" => sbmlRoot,
     "sech" => sech,
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -16,7 +16,7 @@ sbmlNeq(a, b) = !isequal(a, b)
 sbmlLog(x) = sbmlLog(10, x)
 sbmlLog(base, x) = log(base, x)
 
-sbmlPower(x, y) = IfElse.ifelse(y < 0, x^float(y), x^y)  # Todo @mirek: perhaps this needs a method float(y::SBML.MathVal)?
+sbmlPower(x, y) = x^float(y)
 sbmlRoot(x) = sqrt(x)
 sbmlRoot(power, x) = x^(1 / power)
 
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -95,7 +95,7 @@ A dictionary of default constants filled in place of SBML Math constants in the
 function conversion.
 """
 const default_constants =
-    Dict{String,Any}("true" => true, "false" => false, "pi" => pi, "e" => exp(1))
+    Dict{String,Any}("true" => true, "false" => false, "pi" => pi, "e" => exp(1), "exponentiale" => exp(1))
 
 """
 $(TYPEDSIGNATURES)
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -21,4 +21,7 @@
     )
 
     @test isequal(SBML.interpret_math(test), 0.5)
+
+    test = SBML.MathConst("exponentiale")
+    @test isequal(SBML.interpret_math(test), exp(1))
 end
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -11,4 +11,14 @@
     )
 
     @test isequal(SBML.interpret_math(test), 456)
+    
+    test = SBML.MathApply(
+        "power",
+        SBML.Math[
+            SBML.MathVal(2),
+            SBML.MathVal(-1),
+        ],
+    )
+
+    @test isequal(SBML.interpret_math(test), 0.5)
 end


src/interpret.jl
```diff
@@ -16,6 +16,7 @@ sbmlNeq(a, b) = !isequal(a, b)
 sbmlLog(x) = sbmlLog(10, x)
 sbmlLog(base, x) = log(base, x)
 
+sbmlPower(x, y) = IfElse.ifelse(y < 0, x^float(y), x^y)  # Todo @mirek: perhaps this needs a method float(y::SBML.MathVal)?
```
exaexa, line 19: I guess we don't want to throw around `ifElse`s especially if this is gonna get into integration packages.

Option 1 is that we do just
```suggestion
sbmlPower(x, y) = float(x) ^ float(y)
```

Another option would be to just avoid integers getting into the float math -- what about that `interpret_math` would produce floats right away, by defaulting
```
    map_value = (x::SBML.MathVal) -> float(x.val),
```
(i.e. extra float conversion right away?)

paulflang: @isaacsas and I had a discussion in the symbolic-manipulation Julia Slack channel. Personally, I think I don't fully understand what would be the best option.

isaacsas: First a caveat; I don’t understand the full context here as I haven’t looked at where this code is getting used.  My concern is that such conversions could potentially break a purely integer-valued stochastic model depending on where they appear. If they only appear in rate law calculations it shouldn’t be a problem, but if they appear in code corresponding to state change amounts that would be problematic.

exaexa: I wouldn't say that is a big problem (SBML seems to work with all values as with doubles anyway).

Maybe @paulflang can you post which model and what computation actually triggers this? It might be the case that we're converting to floats too late, which would be  a better thing to fix than try to hotpatch all exp calls.

paulflang: @isaacsas : SBMLTk uses `interpret_math` in event triggers, event assignments, rules and rate laws.

@exaexa : see for instance [this rateRule in case 889](https://github.com/sbmlteam/sbml-test-suite/blob/7ab011efe471877987b5d6dcba8cc19a6ff71254/cases/semantic/00889/00889-sbml-l3v2.xml#L43).

isaacsas: > I wouldn't say that is a big problem (SBML seems to work with all values as with doubles anyway).

I’m not sure what you mean by that since SBML is used to represent integer-valued stochastic models, and the example that caused this problem seems to specifically annotate the constants as integer (hence why I discussed with @paulflang whether one could just add a dispatch for the specific case of integer powers).

exaexa: ok I guess the forced conversion is pretty much okay then. I edited my suggestion to a slightly less intrusive form. I wouldn't say we can do much better here without introducing a glue that would be specific to how MTK does this.

isaacsas: What does the Python SBML tooling do here? (Python doesn’t allow integer to negative integer power either right?) Maybe consistency with what is done there is the best choice?

giordano: > Python doesn’t allow integer to negative integer power either right?

I don't think Python cares much about type-stability:
```
In [1]: a = 3

In [2]: b = -4

In [3]: a ** b
Out[3]: 0.012345679012345678

In [4]: 3 ** -4
Out[4]: 0.012345679012345678
```

paulflang: Python seems to allow integer to negative power:
```
Python 3.10.4 (tags/v3.10.4:9d38120, Mar 23 2022, 23:13:41) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> 2**(-1)
0.5
>>> type(-1)
<class 'int'>
```

exaexa: Like, if the additional `if` in the codepath/source for differentiation/integration isn't a problem, I'm perfectly okay with the current version. (with the conservative `if`)

isaacsas: Sorry, I guess it is Numpy integers that don’t/didn’t allow such conversion (I was looking at: https://stackoverflow.com/questions/43287311/why-cant-i-raise-to-a-negative-power-in-numpy as I don’t actually have a Python install…)

exaexa: @paulflang ok I guess we can't do much better here. :]

exaexa (approved):
Please remove the TODO comment.

src/interpret.jl
isaacsas: I don't know the right solution here either, so I'd defer to what you all think is best at this point -- I don't want to push you into too conservative / special cased of a solution. Conversion to `float` does seem cleaner, so perhaps it would make sense to go ahead with that and just revisit this if an example comes up where this actually breaks a model. It may be that such a problem never arises in practice.

PR #220, opened 2022-07-15
remove the debugging leftover

exaexa:
COBREXA is actually able to write a SBML file with this check, so hopefully there are no more :]


giordano (approved):
:innocent:

PR #221, opened 2022-07-15
enable multiple arguments for comparisons

paulflang:
This is needed for instance for case 1215.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -12,6 +12,12 @@ function sbmlPiecewise(args...)
 end
 
 sbmlNeq(a, b) = !isequal(a, b)
+sbmlRelational(fun) = begin
+    function f(x...)
+        all(map(i -> fun(x[i], x[i+1]), 1:length(x)-1))
+    end
+    f
+end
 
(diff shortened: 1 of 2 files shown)


src/interpret.jl
```diff
@@ -12,6 +12,12 @@ function sbmlPiecewise(args...)
 end
 
 sbmlNeq(a, b) = !isequal(a, b)
+sbmlRelational(fun) = begin
+    function f(x...)
+        all(map(i -> fun(x[i], x[i+1]), 1:length(x)-1))
+    end
+    f
+end
```
exaexa, line 20: I think this can be simplified (julia `map` works nicely as `zipWith`):
```suggestion
sbmlRelational(op) =
    (x...) -> all(map(op, x[begin:end-1], x[begin+1:end]))
```

exaexa (commented):
1 suggestion, otherwise lgtm

exaexa (approved):

exaexa:
I'm going to edit this a bit to fix the formatting

PR #224, opened 2022-07-17
implement full species references

### LINKED ISSUE #218: Import species reference metadata
## Target functionality

Allow sensible handling of `Xref` from here:
https://github.com/sbmlteam/sbml-test-suite/issues/87

exaexa:
Closes #218 .

There's a bit of re-juggling of the test file contents because our fav L2 models don't have the SpeciesReferences attributes that are required for L3.


paulflang (commented):
Does a SpeciesReference not need an ID? Consider an assignment rule `A <- 2*SR123`, where `SR123` is a SpeciesReference. How do I know what numeric value to substitute into `SR123`, when SpecieReference does not have an ID?

exaexa:
@paulflang good point, I totally forgot about them. Should be in now.

paulflang (approved):

PR #225, opened 2022-07-18
use the safer Object_createChild SBML methods.

exaexa:
This depends on https://github.com/sbmlteam/libsbml/pull/252 to be merged,
because some of the functions were actually not exposed to the C API.

Creating the members from the parent objects seems safer because all XML
metadata and versioning info gets automatically propagated. Also the code is a
bit shorter.

### DIFF
--- a/src/unitful.jl
+++ b/src/unitful.jl
@@ -101,30 +101,3 @@ string ID.
 """
 unitful(m::Model, val::Tuple{Float64,String}, default_unit::String) =
     unitful(m, val, unitful(m.units[default_unit]))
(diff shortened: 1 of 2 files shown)


giordano (approved):
I agree this pattern is a bit better, when it can be used :+1:

exaexa:
> when it can be used

Patchdozer disagrees. This will be used everywhere. :D

exaexa:
now depends on https://github.com/JuliaPackaging/Yggdrasil/pull/5224

Issue #217, opened 2022-07-14
Check possible silent errors while building SBML structure


exaexa:
As discussed here https://github.com/sbmlteam/libsbml/issues/248, some functions may fail very silently and we should always check the return values to make sure this doesn't happen. Can be done with a specialized call macro or so.

PR #229, opened 2022-09-03
support SBO terms (+ bonuses)

exaexa:
mostly, I shortened the writesbml code and added some error checking. Also, "active objective" is now correctly Maybe (turns out libsbml just doesn't expose the `isSet` API).

### DIFF
--- a/src/structs.jl
+++ b/src/structs.jl
@@ -425,7 +425,7 @@ Base.@kwdef struct Model
     constraints::Vector{Constraint} = Constraint[]
     reactions::Dict{String,Reaction} = Dict()
     objectives::Dict{String,Objective} = Dict()
-    active_objective::String = ""
+    active_objective::Maybe{String} = nothing
     gene_products::Dict{String,GeneProduct} = Dict()
     function_definitions::Dict{String,FunctionDefinition} = Dict()
     events::Dict{String,Event} = Dict()
--- a/src/structs.jl
+++ b/src/structs.jl
@@ -176,6 +176,7 @@ Base.@kwdef struct Parameter
     value::Maybe{Float64} = nothing
     units::Maybe{String} = nothing
     constant::Maybe{Bool} = nothing
+    sbo::Maybe{String} = nothing
 end
 
 """
@@ -194,6 +195,7 @@ Base.@kwdef struct Compartment
     units::Maybe{String} = nothing
     notes::Maybe{String} = nothing
     annotation::Maybe{String} = nothing
+    sbo::Maybe{String} = nothing
 end
 
 """
@@ -235,6 +237,7 @@ Base.@kwdef struct Reaction
     metaid::Maybe{String} = nothing
     notes::Maybe{String} = nothing
     annotation::Maybe{String} = nothing
+    sbo::Maybe{String} = nothing
 end
 
 """
@@ -318,6 +321,7 @@ Base.@kwdef struct Species
     metaid::Maybe{String} = nothing
     notes::Maybe{String} = nothing
     annotation::Maybe{String} = nothing
+    sbo::Maybe{String} = nothing
 end
 
 """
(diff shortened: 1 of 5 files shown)


exaexa:
Closes #217 

Cc @stelmo

src/writesbml.jl
```diff
@@ -82,63 +76,66 @@ function add_rule(model::VPtr, r::Union{AssignmentRule,RateRule})
...
+    set_bool!(parameter_ptr, :Parameter_setConstant, parameter.constant)
+    set_sbo_term!(parameter_ptr, parameter.sbo)
     return parameter_ptr
 end
 
+function set_string!(ptr::VPtr, fn_sym::Symbol, x::Maybe{String})
+    isnothing(x) ||
+        ccall(sbml(fn_sym), Cint, (VPtr, Cstring), ptr, x) == 0 ||
+        error("$fn_sym failed for value `$x' !")
+end
```
giordano, line 111: Nice functions!

src/writesbml.jl
```diff
@@ -82,63 +76,66 @@ function add_rule(model::VPtr, r::Union{AssignmentRule,RateRule})
...
+
+function set_int!(ptr::VPtr, fn_sym::Symbol, x::Maybe{I}) where {I<:Integer}
```
giordano, line 113: If you don't need `I` in the body of the function you can also do `Maybe{<:Integer}`

src/writesbml.jl
```diff
@@ -777,6 +508,13 @@ function _create_doc(mdl::Model)::VPtr
     return doc
 end
 
+"""
+$(TYPEDSIGNATURES)
+
+Write the SBML structure in `mdl` to a file `fn`.
+
+To write the XML to a string, use the other overload.
```
giordano, line 516: Maybe make it more explicit like

> To write the XML to a string, use `writeSBML(mdl::Model)`

?  Likewise for the other docstring.

test/loadmodels.jl
```diff
@@ -376,7 +376,7 @@ end
     @test m.substance_units == "substance"
     @test m.time_units == "second"
     @test m.volume_units == "volume"
-    @test m.active_objective == ""
+    @test m.active_objective == nothing
```
giordano, line 379:
```suggestion
    @test isnothing(m.active_objective)
```

test/loadmodels.jl
```diff
@@ -388,15 +388,19 @@ end
     @test isnothing(m.substance_units)
     @test isnothing(m.time_units)
     @test isnothing(m.volume_units)
-    @test m.active_objective == ""
+    @test m.active_objective == nothing
```
giordano, line 391:
```suggestion
    @test isnothing(m.active_objective)
```

src/structs.jl
```diff
@@ -420,7 +425,7 @@ Base.@kwdef struct Model
     constraints::Vector{Constraint} = Constraint[]
     reactions::Dict{String,Reaction} = Dict()
     objectives::Dict{String,Objective} = Dict()
-    active_objective::String = ""
+    active_objective::Maybe{String} = nothing
```
giordano, line 428: Note that I originally [had this as `Maybe{String}`](https://github.com/LCSB-BioCore/SBML.jl/commit/a3461be3d566e1696ddc08f45115b5a85ea72b7f), but then [changed it to `String`](https://github.com/LCSB-BioCore/SBML.jl/commit/f385a0da62e4f05de7cba9255cd79f4cc2579903) because libsbml treats the active objective ID as an empty string, instead of `NULL`, when there is no active objective.  I think I had problems with the roundtrip checks because of this.

test/loadmodels.jl
exaexa: uh whoops. :D

src/structs.jl
exaexa: yeah the `FbcModelPlugin_isSetActiveObjective` is missing in the C API, so now I check the string emptiness. Otherwise it's the same thing as with the other emptiable strings. I'll eventually make a PR fix to libsbml.

src/writesbml.jl
exaexa: true, thanks

src/structs.jl
giordano: Ok, sounds good!

giordano (approved):

PR #230, opened 2022-09-08
fix sbmlRelational error with `Num` arguments

paulflang:


exaexa:
Merging into a temp branch, will fix there.

PR #231, opened 2022-09-09
cleanup&fix the sbmlRelational patch

exaexa:


exaexa:
cc @paulflang

giordano (commented):
Looks ok to me, but if #230 is a bugfix, shouldn't it also come with a test?

exaexa:
Ah true.

giordano (approved):

PR #233, opened 2022-09-11
implement the special handling for avogadro constant

exaexa:
@paulflang can you check if this would do the trick as discussed on slack?

### DIFF
--- a/src/structs.jl
+++ b/src/structs.jl
@@ -139,7 +139,7 @@ end
 """
 $(TYPEDEF)
 
-A special value representing the avogadro constant (which is a special named
+A special value representing the Avogadro constant (which is a special named
 value in SBML).
 
 # Fields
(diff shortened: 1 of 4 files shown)


exaexa:
PS can you recommend a SBML test case that would contain this thing?

paulflang:
Thanks, looks good to me. I cannot recommend a particular test case. I guess any of those you grepped in you Slack message will be fine.

src/structs.jl
```diff
@@ -139,6 +139,19 @@ end
 """
 $(TYPEDEF)
 
+A special value representing the avogadro constant (which is a special named
```
giordano, line 142:
```suggestion
A special value representing the Avogadro constant (which is a special named
```
:wink:

giordano (approved):
Lovely

src/structs.jl
exaexa: image
tho.

PR #236, opened 2022-10-03
`use_value_from_trigger_time` is semantically a Bool

exaexa:
might be breaking, but technically it only breaks broken code.... :D

### DIFF
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -659,7 +659,7 @@ function get_model(mdl::VPtr)::SBML.Model
                     Cint,
                     (VPtr,),
                     ev,
-                ),
+                ) != 0,
                 name = get_optional_string(ev, :Event_getName),
                 trigger,
                 event_assignments,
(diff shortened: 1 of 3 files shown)


src/readsbml.jl
```diff
@@ -659,7 +659,7 @@ function get_model(mdl::VPtr)::SBML.Model
                     Cint,
```
giordano, line 659: If we're ***sure*** this is a boolean value, this can also be
```suggestion
                    Bool,
```
and you don't need to add `!= 0` below, with the caveat that `1` is `true`, and all the rest is `false`:
```julia
julia> f = @cfunction identity Cint (Cint,)
Ptr{Nothing} @0x00007fa9afb184d0

julia> ccall(f, Cint, (Cint,), 2)
2

julia> ccall(f, Bool, (Cint,), 2)
false

julia> ccall(f, Bool, (Cint,), 1)
true

julia> ccall(f, Bool, (Cint,), 0)
false
```
which might be a bit counter-intuitive.

exaexa:
```julia
julia> ccall(f, Bool, (Cint,), 6)
false
```

ok well now this is scary. I'd use it as-is because that is what we have in the get_optional_bool above, and also the function signatures should match.

PR #238, opened 2023-01-22
add convert_promote_expand

paulflang:
The initial assignment may contain parameters. If these parameters shall be optimised, the `convert_simplify_math` converter must not be used. Instead, the `convert_promote_expand` converter from this PR should be used. I could put it into SBMLtk, but I think it better fits in SBML.jl, cause this is the package that actually does the conversion.


exaexa (commented):
sounds good to me; I'd only go for `convert_promote_params_and_expand` or something so that it's clear that it doesn't promote some expansion or whatever.

exaexa (approved):
OK with me; if you want to have this released pls bump version in Project.toml to 1.4.0 before merging.

PR #243, opened 2023-01-30
make throwing SBML load errors configurable

exaexa:


giordano (approved):
Nice!   Should be tested?

anandijain:
is there one where i can just not see anything or pass an IO to redirect it?
when I run on all the biomodels, im going in expecting a bunch of warnings, but I don't need to see any of them

anandijain:
is there one where i can just not see anything or pass an IO to redirect it?
when I run on all the biomodels, im going in expecting a bunch of warnings, but I don't need to see any of them

exaexa:
> Nice! Should be tested?

Good question. I'd need to add a small broken model to test this properly, will add an issue and do it sometimes in the future.

anandijain:
we should add that to the docs, i got the impression the only options were warn or fatal lol

exaexa:
> we should add that to the docs, i got the impression the only options were warn or fatal lol

but it's in the docstrings?

anandijain:
mirek! i cant read! didn't you know

PR #247, opened 2023-02-15
support more CVTerms and annotations

exaexa:
Mainly a dependency of https://github.com/LCSB-BioCore/COBREXA.jl/pull/750 .

### DIFF
--- a/src/structs.jl
+++ b/src/structs.jl
@@ -208,6 +208,8 @@ Base.@kwdef struct Parameter
     units::Maybe{String} = nothing
     constant::Maybe{Bool} = nothing
     metaid::Maybe{String} = nothing
+    notes::Maybe{String} = nothing
+    annotation::Maybe{String} = nothing
     sbo::Maybe{String} = nothing
     cv_terms::Vector{CVTerm} = []
 end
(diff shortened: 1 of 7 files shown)


src/structs.jl
```diff
@@ -479,4 +484,6 @@ Base.@kwdef struct Model
     volume_units::Maybe{String} = nothing
     notes::Maybe{String} = nothing
     annotation::Maybe{String} = nothing
+    sbo::Maybe{String} = nothing
+    cv_terms::Vector{CVTerm} = []
```
giordano, line 488: You may want to add these two fields to the tests in https://github.com/LCSB-BioCore/SBML.jl/blob/0eb0a446b62b5072e851d61fcab1e2e0c66b8124/test/writemodels.jl#L58

src/structs.jl
```diff
@@ -208,6 +208,8 @@ Base.@kwdef struct Parameter
     units::Maybe{String} = nothing
     constant::Maybe{Bool} = nothing
     metaid::Maybe{String} = nothing
+    notes::Maybe{String} = nothing
+    annotation::Maybe{String} = nothing
```
giordano, line 212: I was going to suggest to add `Parameter` to the list of structs with annotations in the tests, but somehow it was already there: https://github.com/LCSB-BioCore/SBML.jl/blob/0eb0a446b62b5072e851d61fcab1e2e0c66b8124/test/common.jl#L25-L29 :sweat_smile:

giordano: We could actually use the same `==` definition (the one which checks for `annotations` and `notes`) for all structs, but I think it'll be slightly slower for structs without annotations.  Performance is probably not a huge concern here though.

src/structs.jl
exaexa: I actually systematized it to whole model, that way it's much easier.

giordano: Uh, yeah, good idea.  I think I added each field manually there before having the smarter custom `==` method for the other structs.

giordano (approved):

PR #250, opened 2023-07-15
add sbmlPower dispatch

paulflang:
This fixes the following bug I saw:
```
ERROR: ArgumentError: Cannot convert Sym to AbstractFloat since Sym is symbolic and AbstractFloat is concrete. Use `substitute` to replace the symbolic unwraps.
Stacktrace:
  [1] AbstractFloat(x::SymbolicUtils.BasicSymbolic{Real})
    @ Symbolics ~/.julia/packages/Symbolics/k84uz/src/Symbolics.jl:146
  [2] float(x::SymbolicUtils.BasicSymbolic{Real})
    @ Base ./float.jl:294
  [3] sbmlPower(x::SymbolicUtils.BasicSymbolic{Real}, y::SymbolicUtils.BasicSymbolic{Real})
    @ SBML ~/.julia/packages/SBML/GYWJL/src/interpret.jl:24
```

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -21,7 +21,8 @@ end
 sbmlLog(x) = sbmlLog(10, x)
 sbmlLog(base, x) = log(base, x)
 
-sbmlPower(x, y) = x^float(y)
+sbmlPower(x, y::Int) = x^float(y)
+sbmlPower(x, y) = x^y
 sbmlRoot(x) = sqrt(x)
 sbmlRoot(power, x) = x^(1 / power)
 
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -21,7 +21,7 @@ end
 sbmlLog(x) = sbmlLog(10, x)
 sbmlLog(base, x) = log(base, x)
 
-sbmlPower(x, y::Int) = x^float(y)
+sbmlPower(x::Int, y::Int) = x^float(y)
 sbmlPower(x, y) = x^y
 sbmlRoot(x) = sqrt(x)
 sbmlRoot(power, x) = x^(1 / power)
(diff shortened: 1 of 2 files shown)


src/interpret.jl
```diff
@@ -21,7 +21,8 @@ end
 sbmlLog(x) = sbmlLog(10, x)
 sbmlLog(base, x) = log(base, x)
 
-sbmlPower(x, y) = x^float(y)
+sbmlPower(x, y::Int) = x^float(y)
```
exaexa, line 24: this might be extended to y::Number or at least y::Real, right?
```suggestion
sbmlPower(x, y::Real) = x^float(y)
```

exaexa (commented):
I added one extra idea, otherwise I guess good to go. Thanks!

If you want this released please bump the patch version.

src/interpret.jl
paulflang: I don't think this is necessary, cause it is only negative Ints in the exponent that cause trouble. Or do you know another case where simply x^y does not work?

exaexa: ah yeah, actually what would break there with the negative exponent? can we omit the float() conversion altogether?

I guess it would be much better to actually find and document the reason for the extra `float` there; if the code is like this we'll break it with the next bugfix of something else :D

src/interpret.jl
```diff
@@ -21,7 +21,7 @@ end
 sbmlLog(x) = sbmlLog(10, x)
 sbmlLog(base, x) = log(base, x)
 
-sbmlPower(x, y::Int) = x^float(y)
+sbmlPower(x::Int, y::Int) = x^float(y)
```
exaexa, line 24: this still doesn't cover ugly cases like ` 0x0005 ^ Int32(-5)`...will fix :D
```suggestion
sbmlPower(x::Integer, y::Integer) = x^float(y)
```

exaexa (changes requested):

exaexa (approved):
good, thx

PR #253, opened 2023-07-18
use IfElse.jl for Julia 1.6"

paulflang:
I could imagine that  #251 fails on Julia 1.6, because it requires IfElse.jl. Let's see if CI passes for this.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -7,7 +7,7 @@ function sbmlPiecewise(args...)
     elseif length(args) >= 3
         IfElse.ifelse(args[2], args[1], sbmlPiecewise(args[3:end]...))
     else
-        throw(DomainError(args, "malformed piecewise SBML function"))
+        IfElse.ifelse(args[2], args[1], NaN)
     end
 end
 


src/interpret.jl
```diff
@@ -7,7 +7,7 @@ function sbmlPiecewise(args...)
     elseif length(args) >= 3
         IfElse.ifelse(args[2], args[1], sbmlPiecewise(args[3:end]...))
     else
-        throw(DomainError(args, "malformed piecewise SBML function"))
+        IfElse.ifelse(args[2], args[1], NaN)
```
exaexa, line 10: this could work but we need an user-supplied "bottom" value instead of the NaN here.

Does IfElse have anything like that?

exaexa (changes requested):

src/interpret.jl
paulflang: Not that I am aware of.

exaexa:
closing for inactivity; please feel free to reopen if there's some good replacement for the NaN.

Btw might be useful to also recommend people to override the sbmlPiecewise with watever they like, which could be handled by the docs, is that OK?

PR #256, opened 2023-10-09
bump documenter version, use doc pkg cache

exaexa:


.github/workflows/docs.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
```
giordano, line 17:
```suggestion
      - uses: actions/checkout@v4
```
Related, it'd be good to add dependatbot to automate this: https://github.com/JuliaPhysics/Measurements.jl/blob/21cf93c7dc25b7a1b14d5b877bc8f4884599ff5e/.github/dependabot.yml

giordano (approved):
Looks good to me if it works :+1:

.github/workflows/docs.yml
exaexa: hm true. I actually thought I'd make a common repo of all CI scriptage and update from there, dependabit might work too. Will have a look asap, thanks.

exaexa: (applied manually)

PR #259, opened 2023-11-13
Add support for specie conversionFactor

### LINKED ISSUE #257: Add support for species conversion factor
## Target functionality

With SBML level 3 species conversion factors are allowed, but I do not think SBML.jl supports these yet (I have not found it in the `SBML.Model` struct). The code I would want to be able to run is;

```
using SBML
using Downloads
path = Downloads.download("https://raw.githubusercontent.com/sbmlteam/sbml-test-suite/release/cases/semantic/00976/00976-sbml-l3v2.xml", joinpath(@__DIR__, "model.xml"))
model = readSBML(path)
model.species["S1"].conversion_factor
```

## Desired output

In case a species (like `S1` here) has a conversion factor I would like the following output :

```
S1conv
```

While for species without conversion factor in the `SBML.Species` struct `nothing` should be returned.

## Optional: Suggestions for implementation

I think this can be implemented by adding a `conversion_factor` field to `SBML.Species` , and a check that a given conversion factor is an identifier of a parameter in the model.

The description of conversion factor can be found under the heading *The conversionFactor attribute in SBML Level 3* in the libSBML manual https://synonym.caltech.edu/software/libsbml/5.18.0/docs/formatted/python-api/classlibsbml_1_1_species.html.

sebapersson:
Closes #257 

With this implementation the PEtab.jl SBML importer passes the SBML tests with specie conversion factor. Also, I do not encounter any problem on the other test cases.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SBML"
 uuid = "e5567a89-2604-4b09-9718-f5f78e97c3bb"
 authors = ["Mirek Kratochvil <miroslav.kratochvil@uni.lu>", "LCSB R3 team <lcsb-r3@uni.lu>"]
-version = "1.5.0"
+version = "1.5.1"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -412,9 +412,9 @@ function get_model(mdl::VPtr)::SBML.Model
                 ccall(sbml(:Species_getInitialConcentration), Cdouble, (VPtr,), sp)
             end,
             substance_units = get_optional_string(sp, :Species_getSubstanceUnits),
-	    conversion_factor = if (
+            conversion_factor = if (
                 ccall(sbml(:Species_getConversionFactor), Cstring, (VPtr,), sp) != C_NULL
-            )   
+            )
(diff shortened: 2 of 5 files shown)


exaexa (commented):
Looks OK to me, thanks a lot! I'll need to add a writeSBML equivalent, but that should be it. Tests are running now, if formatting fails don't worry about it.

We should also have test coverage -- can you please send me a link (or ideally SBML testcase number :) ) to a model that would require this functionality?

sebapersson:
Thanks! A test case is semantic 00976 can be found here https://github.com/sbmlteam/sbml-test-suite/tree/release/cases/semantic/00976

sebapersson:
@exaexa any status on this?

I could try to also add a test and support for writeSBML, but for the test, I do not really understand the structure of the loadmodels.jl test file

exaexa:
Hi, yes this is in the queue, I will finish the tests and there is no reason for this to not get merged. Sorry for delays, I'm off work these days and don't have much time left for solving stuff.

exaexa:
@sebapersson I guess we're fixed; I'll release this ASAP once it's merged. You can set your dependencies to SBML v1.5.1 to properly depend on species conversion factors availablility. Thanks again!

exaexa:
Ref: https://github.com/JuliaRegistries/General/pull/95464

PR #260, opened 2023-12-11
add support for groups extension

exaexa:


exaexa:
(@giordano this might take some time to debug but any preliminary comments very welcome :) )

exaexa:
note to self: TODO improve coverage with adding sbml testcases

src/structs.jl
```diff
@@ -450,6 +450,41 @@ end
...
+    sbo::Maybe{String} = nothing
+    cv_terms::Vector{CVTerm} = []
+end
+
+"""
+$(TYPEDEF)
+
 Structure that collects the model-related data. Contains `parameters`, `units`,
 `compartments`, `species` and `reactions` and `gene_products`, and additional
 `notes` and `annotation` (also present internally in some of the data fields).
```
giordano, line 490: This list is getting out of date.  But at least all fields are automatically listed in the section below.  Not sure it's worth naming the fields here?

giordano (approved):
Looks good to me!

exaexa:
@mihai-sysbio btw if you could have a quick look at `src/structs.jl` and confirm that it would be sufficient to support the StdGem/YeastGem usecases it would be great :] Thanks!

stelmo (approved):
SBML 😨

src/structs.jl
```diff
@@ -450,6 +450,41 @@ end
 """
 $(TYPEDEF)
 
+# Fields
+$(TYPEDFIELDS)
+"""
+Base.@kwdef struct Member
```
giordano, line 456: Do these new structs need to be added to https://github.com/LCSB-BioCore/SBML.jl/blob/64ec00b69130a70cf75775792e76c7db6003b1c3/test/common.jl#L22-L33?

exaexa: good point thx, done now

src/structs.jl
exaexa: solved now

mihai-sysbio:
> @mihai-sysbio btw if you could have a quick look at `src/structs.jl` and confirm that it would be sufficient to support the StdGem/YeastGem usecases it would be great :] Thanks!

Sorry I was slow to reply. It looks like a very flexible structure is now I place 👍🏻  I must admit I haven't looked into other SBMLs and how the structure these groups, so I would have guessed the Members of a Group would have fewer attributes.

PR #282, opened 2026-09-12
Harden i686 libsbml ccall/string handling

ChrisRackauckas-Claude:
## Summary
- SciML/SBMLToolkitTestSuite x86 fails precompiling SBML with `ArgumentError: cannot convert NULL to string` in `get_model`.
- Treat libsbml predicate returns as `Cint`, null-check C strings/pointers before `unsafe_string`, and skip the precompile warmup on 32-bit.

## Test plan
- [ ] `using SBML` on i686 no longer aborts at precompile
- [ ] Existing 64-bit tests still pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/SBML.jl
+++ b/src/SBML.jl
@@ -35,6 +35,10 @@ export set_level_and_version,
     libsbml_convert, convert_simplify_math, convert_promotelocals_expandfuns
 
 # Read a file at precompile time, to improve time-to-first `readSBML`.
-writeSBML(readSBML(joinpath(@__DIR__, "..", "test", "data", "Dasgupta2020-written.xml")))
+# Skip on 32-bit: SBML_jll/libsbml has returned NULL C strings during model
+# walks on i686 (ArgumentError: cannot convert NULL to string).
+if Sys.WORD_SIZE == 64
+    writeSBML(readSBML(joinpath(@__DIR__, "..", "test", "data", "Dasgupta2020-written.xml")))
+end
 
 end # module
--- a/src/readsbml.jl
+++ b/src/readsbml.jl
@@ -143,7 +143,7 @@ $(TYPEDSIGNATURES)
 Helper for converting XML that is not represented by SBML structures to String.
 """
 function get_string_from_xmlnode(xmlnode::VPtr)::String
-    if ccall(sbml(:XMLNode_isText), Bool, (VPtr,), xmlnode)
+    if ccall(sbml(:XMLNode_isText), Cint, (VPtr,), xmlnode) != 0
         str_ptr = ccall(sbml(:XMLNode_getCharacters), Cstring, (VPtr,), xmlnode)
         str_ptr == C_NULL ? "" : unsafe_string(str_ptr)
     else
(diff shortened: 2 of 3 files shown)


src/SBML.jl
```diff
@@ -35,6 +35,10 @@ export set_level_and_version,
     libsbml_convert, convert_simplify_math, convert_promotelocals_expandfuns
 
 # Read a file at precompile time, to improve time-to-first `readSBML`.
-writeSBML(readSBML(joinpath(@__DIR__, "..", "test", "data", "Dasgupta2020-written.xml")))
+# Skip on 32-bit: SBML_jll/libsbml has returned NULL C strings during model
+# walks on i686 (ArgumentError: cannot convert NULL to string).
```
exaexa: "walks" ?

src/readsbml.jl
```diff
@@ -143,7 +143,7 @@ $(TYPEDSIGNATURES)
 Helper for converting XML that is not represented by SBML structures to String.
 """
 function get_string_from_xmlnode(xmlnode::VPtr)::String
-    if ccall(sbml(:XMLNode_isText), Bool, (VPtr,), xmlnode)
+    if ccall(sbml(:XMLNode_isText), Cint, (VPtr,), xmlnode) != 0
```
exaexa: this looks valid, C API actually returns `int`s unlike the C++ one.

exaexa:
I'll have to split this to smaller pieces. The `CInt` vs `Bool` problem is valid (what is the fix coverage at this point?), but some parts of the patch are questionable.

- is there a valid scenario when the `Model_getInitialAssignment` returns actual null (except if the SBML structure is internally broken AF)? If so, we need a testcase. The same method is used to do literally all processing of SBML ListOf, fixing this single instance doesn't seem to make any systematic sense.
- why disable the preload on `i686` if this should have actually fixed the `i686` in the first place??
- what was the actual point where the NULL string has appeared? Is there a test case from SBML testsuite?

(Again, kindly asking for human review of PR sanity before submitting.)

src/readsbml.jl
```diff
@@ -645,9 +645,10 @@ function get_model(mdl::VPtr)::SBML.Model
     num_ias = ccall(sbml(:Model_getNumInitialAssignments), Cuint, (VPtr,), mdl)
     for i = 0:(num_ias-1)
         ia = ccall(sbml(:Model_getInitialAssignment), VPtr, (VPtr, Cuint), mdl, i)
+        ia == C_NULL && continue
         sym = ccall(sbml(:InitialAssignment_getSymbol), Cstring, (VPtr,), ia)
         math_ptr = ccall(sbml(:InitialAssignment_getMath), VPtr, (VPtr,), ia)
-        if math_ptr != C_NULL
+        if math_ptr != C_NULL && sym != C_NULL
```
exaexa, line 651: @giordano this would cause the loading to silently ignore contents of invalid SBML instead of complaining loudly, do we want that kind of behavior?

exaexa: (other checks do similarly)

giordano: Loud, loud!

exaexa:
@ChrisRackauckas can you please link the exact log from your x86/i686 runs that show what _exactly_ is broken there? It looks quite unlikely to me that the crash would be caused solely by the string handling (which is safe this way). I guess it's more about the `bool`s but can't be sure how much of that got hidden in slop here.

ChrisRackauckas:
https://github.com/SciML/SBMLToolkit.jl/actions/runs/34130631830/job/101773173225#step:12:1378

It's just in the CI, if you enable the CI you'd see it.

exaexa:
some of the fixes landed in #286 , we'll try to add the i686 tests asap; sounds like a good idea.

PR #283, opened 2026-09-19
fix readsbml Bool vs C int confusion

exaexa:
Originally proposed in #282 , thanks @ChrisRackauckas .

Actually grepped all the Bools and compared to docs, so this should be complete.


src/readsbml.jl
```diff
@@ -143,7 +143,7 @@ $(TYPEDSIGNATURES)
 Helper for converting XML that is not represented by SBML structures to String.
 """
 function get_string_from_xmlnode(xmlnode::VPtr)::String
-    if ccall(sbml(:XMLNode_isText), Bool, (VPtr,), xmlnode)
+    if ccall(sbml(:XMLNode_isText), CInt, (VPtr,), xmlnode) != 0
```
giordano:
```suggestion
    if ccall(sbml(:XMLNode_isText), Cint, (VPtr,), xmlnode) != 0
```
and everywhere below.

src/readsbml.jl
```diff
@@ -572,7 +572,7 @@ function get_model(mdl::VPtr)::SBML.Model
         end
 
         # explicit reversible flag (defaults to true in SBML)
-        reversible = Bool(ccall(sbml(:Reaction_getReversible), Cint, (VPtr,), re))
+        reversible = ccall(sbml(:Reaction_getReversible), Cint, (VPtr,), re) != 0
```
giordano, line 575: This was unnecessary?

exaexa:
Note for self, the tests need serious fixing

- JuliaFormatter does a weird format of a lambda in one case
- we finally need a mirror of biomodels contents :D
- `Dict` hash changed in Julia 1.13 and the SBML tests stupidly depended on it, so there should be either `SortedDict` or `sort` in tests

src/readsbml.jl
exaexa: oh whoops thanks. :D

src/readsbml.jl
exaexa: This is the same type of conversion so I thought it should be consistent

giordano: I don't think this conversion is problematic. The problem may have been the automatic conversion done by ccall, perhaps there's an ABI mismatch between the C bool and the int, but without knowing what was the actual issue it's hard to tell.

ChrisRackauckas:
It might be good to start by enabling the 32-bit tests?

src/readsbml.jl
exaexa: yeah as far as I got it, the problem happens on 32b where the bool somehow really misaligns/mismatches and stuff explodes

giordano: Yes, but that's probably only during the automatic `ccall` conversion, that really requires getting the ABI of the C library right.  Converting "manually" should be fine, otherwise the entire 32-bit build would be unusable.

exaexa: Ahh, I answered the wrong question. I converted this because the Int to Bool conversion in the rest of the code is now done via `!= 0` so having an `Bool(...)` here instead seemed inconsistent; and this method sounds a bit more reliable to me w.r.t. what C considers `true`.

giordano: Yes, converting a plain `Int32` to plain `Bool` should always work, otherwise there would be bigger issues.  Inside a `ccall`, there are some automatic conversions done based on the ABI of the specified types.  In general we should follow closely what a C library does.  I'm pretty sure I wrote the `ccall(..., Bool, ...)` because I thought that was smart, which does work...as long as the ABI of a boolean on the platform matches that of `Int32`, which is actually what the library uses

PR #285, opened 2026-09-22
use the new mirror for downloading some of the models

exaexa:
see https://github.com/COBREXA/models-mirror


exaexa:
(ingested into #286)

